PR #15, opened 2018-06-05
Setup Travis CI and various badges

tkuhlman:


.travis.yml
```diff
@@ -0,0 +1,19 @@
...
+  - go get -u github.com/golang/dep/...
+  - dep ensure
+  - go get github.com/mattn/goveralls
+
+script:
+  - go test -v -race ./...
+  - go test -coverprofile=coverage.txt ./...
+
+after_success:
+  - goveralls -coverprofile=coverage.txt -service=travis-ci
```
scniro: 💯

.travis.yml
```diff
@@ -0,0 +1,19 @@
+language: go
+go:
+  - 1.10.2
+
+services:
+  - docker
+sudo: required
```
scniro: have you tried `false` here? I know when true travis spins up a whole VM vs. a container if false. I know it's needed for some stuff so I'd be curious what the need is

scniro: Looks like we can go containers, https://docs.travis-ci.com/user/reference/overview/#Virtualization-environments

tkuhlman: good point I don't know if it is needed at all or not. I'll try without.

scniro: Sweet, `sudo: false` should be the only change here and you'll notice a huge increase in build time

PR #21, opened 2018-06-25
PENG-1842 dereferencing json logic

ikenticus:


paas-api-pr-manager:

* [PENG-1842](https://jira.gannett.com/browse/PENG-1842): Extend the jstransform jsonschema package to fully handle jsonRefs
```
The jsonschema package can read in JSON schema files and expose them to other tooling but it currently only handles simple file refs at the root level. Extend it to more completely support full JSON schema refs.

Accetance Criteria:
 * A ref anywhere in the file is handled
 * refs to other sections of the document work fine
```

jsonschema/reference.go
```diff
@@ -0,0 +1,168 @@
...
+					top = top.([]interface{})[intKey]
+				} else {
+					top = top.(map[string]interface{})[item]
+				}
+			} else {
+				targetRef, err := buildReference(schemaPath, data, ref.Target)
+				if err != nil {
+					return dj, input, fmt.Errorf("unable to build reference from %s: %v", ref.Target, err)
+				}
+				targetKeys := reflect.ValueOf(targetRef).MapKeys()
```
tkuhlman: It is a bit funky to do reflection to get a length, I think better would be for buildReference to not return an empty interface.

jsonschema/schema.go
```diff
@@ -67,11 +67,24 @@ func SchemaFromFile(schemaPath string, oneOfType string) (*Schema, error) {
 		return nil, fmt.Errorf("failed to initialize schema validator: %v", err)
 	}
 
+	var dj schemaJSON // deferenced JSON
+	dj, data, err = Dereference(schemaPath, data)
+	if err != nil {
+		return nil, fmt.Errorf("failed to Dereference Schema: %v", err)
+	}
+
 	var sj schemaJSON
```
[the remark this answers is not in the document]
ikenticus: Initially, I did not want to return schemaJSON at all, but I was unable to figure out how to remove all schemaJSON from the schema.go (and the generate/walk)

jsonschema/reference.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+		json.Unmarshal([]byte(data), &source)
+	}
+	return parseReference(source, strings.Split(target[1], "/")[1:]), nil
+}
+
+// parseReference recursively parses the given reference path
+func parseReference(source interface{}, refPaths []string) interface{} {
+	if len(refPaths) > 1 {
+		return parseReference(source.(map[string]interface{})[refPaths[0]], refPaths[1:])
```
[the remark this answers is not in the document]
ikenticus: Yeah, had thought of writing up a test for this --- just wasn't sure if I needed to test all the "non-public" (I forget the term) functions.  I originally had intended to write a test for this and so I will...

jsonschema/reference.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+// jsonRef represents a JSON Reference source and targetRef
+type jsonRef struct {
+	Source []string
+	Target string
+}
+
+// Dereference parse JSON string and replaces all $ref with the referenced data.
+func Dereference(schemaPath string, input []byte) (schemaJSON, []byte, error) {
```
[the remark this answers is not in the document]
ikenticus: Originally, it operated purely on []byte.  The schemaJSON being returned was just my "hack" of solving the failing schema/generate/walk tests --- I'll revert it to my original without the schemaJSON and try to figure out a better way to resolve the other failing tests.

jsonschema/reference.go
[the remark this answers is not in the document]
tkuhlman: There isn't a hard rule for unit tests when the functions like this are covered by other tests it is just the combination of recursion and empty interfaces that made me think it would benefit from that.

jsonschema/reference.go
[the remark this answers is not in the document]
tkuhlman: Yeah I'm curious what the tests are that are failing. If this returns a bunch of JSON as a []byte without references and the rest of the code can't handle it then I suspect there is a bug there.

jsonschema/reference.go
```diff
@@ -0,0 +1,161 @@
...
+	refs, err := walkInterface(data, []string{}, []jsonRef{})
+	if err != nil {
+		return input, fmt.Errorf("unable to walk interface %s: %v", schemaPath, err)
+	}
+
+	for _, ref := range refs {
+		top := data
+		for i, item := range ref.Source {
+			if isTop && (ref.Source[0] == "allOf" || ref.Source[0] == "oneOf") {
+				continue // do not dereference top-level allOr and oneOf
```
tkuhlman: allOf not allOr

jsonschema/reference.go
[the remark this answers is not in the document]
tkuhlman: Nice test, you should still check the type assertion though, it is quite unlikely to fail but if it does it will panic and so better to check it. Same for the other type assertion below. Something like
`sourceMap, ok := source.(map[string]interface{})
if !ok {//error}
`

jsonschema/reference.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+	}
+
+	return json.Marshal(data)
+}
+
+// walkInterface traverses the map[string]interface{} to located json references
+func walkInterface(node interface{}, source []string, refs []jsonRef) ([]jsonRef, error) {
+	var err error
+	for key, val := range node.(map[string]interface{}) {
```
tkuhlman: Another unchecked type assertion here, like I said below you should either check it or just make node be a map[string]interface{} in the function arguments.

jsonschema/schema.go
```diff
@@ -10,27 +10,16 @@ import (
...
-// schemaJSON represents the basic supported structure of a JSON Schema file
-type schemaJSON struct {
-	Instance
-
-	Schema      string      `json:"$schema"`
-	Description string      `json:"description,omitempty"`
-	AllOf       []schemaRef `json:"allOf,omitempty"`
-	OneOf       []schemaRef `json:"oneOf,omitempty"`
-	Required    []string    `json:"required,omitempty"`
+	Ref         string                     `json:"$ref,omitempty"`
```
tkuhlman: I like working toward more complete compliance with the full JSON schema spec. Clearly what was there before was a bit of a bare minimum, nice improvement.

jsonschema/reference.go
tkuhlman: By doing this the former handling of allOf and oneOf doesn't change. That makes sense to do as an interim but I'm not sure it makes sense to do long term. Can you follow up this merging with a new ticket to investigate handling allOf/oneOf in the same way as all other refs? That would allow us to do things like export this function it could be usable by anyone importing the library as a standalone dereference utility.

jsonschema/reference.go
```diff
@@ -0,0 +1,161 @@
...
+				targetRef, err := buildReference(schemaPath, data, ref.Target)
+				if err != nil {
+					return input, fmt.Errorf("unable to build reference from %s: %v", ref.Target, err)
+				}
+				targetKeys := reflect.ValueOf(targetRef).MapKeys()
+				if len(targetKeys) > 1 {
+					// assuming integer item is slice[index] instead of map[string]
+					if intKey, err := strconv.Atoi(item); err == nil {
+						top.([]interface{})[intKey] = targetRef
+					} else {
```
tkuhlman: I would be really curious about running gocyclo on this to find the cyclomatic complexity. I'm out of time to think harder about how to simplify it but it looks really complex on first glance.

ikenticus:
```
$ gocyclo -top 10 ./jsonschema/
18 jsonschema TestWalkJSONSchema jsonschema/walk_test.go:20:1
14 jsonschema dereference jsonschema/reference.go:23:1
14 jsonschema TestWalkJSONSchemaRaw jsonschema/walk_test.go:292:1
12 jsonschema SchemaFromFile jsonschema/schema.go:49:1
11 jsonschema TestSchemaFromFile jsonschema/schema_test.go:82:1
11 jsonschema TestDereference jsonschema/reference_test.go:14:1
11 jsonschema walkInstance jsonschema/walk.go:52:1
10 jsonschema buildReference jsonschema/reference.go:113:1
10 jsonschema walkInterface jsonschema/reference.go:80:1
6 jsonschema TestValidate jsonschema/validator_test.go:8:1
```

jsonschema/reference.go
ikenticus: I don't think I can make node a map[string]interface{} due to it changing between string/slice/map during recursion, so will add in the assertion check

jsonschema/reference.go
ikenticus: Agreed, I had thought of extending the allOf/oneOf in the jstransform logic, but even more complex than the scope of this ticket (allOf may be less complex than oneOf).  I will create a follow up ticket.

ikenticus: Adding PENG-4337 to non-sprint, most likely to fall over into the next sprint.

jsonschema/reference.go
ikenticus: Was not able to add the assertion check INSIDE parseReference due to the multi/single-level conflict, so I added before the parseReference call.

jsonschema/reference.go
```diff
@@ -91,6 +91,10 @@ func walkInterface(node interface{}, source []string, refs []jsonRef) ([]jsonRef
 				})
 			}
 		case reflect.Slice:
+			valMap, ok := val.([]interface{})
+			if !ok {
+				return nil, fmt.Errorf("failed assertion of val: %q", valMap)
+			}
 			for i, item := range val.([]interface{}) {
```
tkuhlman: Since you now have valMap you can put it into the ragne statement.

jsonschema/reference.go
```diff
@@ -100,6 +104,10 @@ func walkInterface(node interface{}, source []string, refs []jsonRef) ([]jsonRef
 				}
 			}
 		case reflect.Map:
+			nodeMap, ok := node.(map[string]interface{})
+			if !ok {
+				return nil, fmt.Errorf("failed assertion of node: %q", nodeMap)
+			}
 			refs, err = walkInterface(node.(map[string]interface{})[key], append(source, key), refs)
```
tkuhlman: you can use nodeMap now rather than redoing the type assertion

jsonschema/reference.go
tkuhlman: Thanks for running that, looks like I was right, other than some tests that I wrote it is the most complicated. More motivation to circle back and try and clean it all up later.

jsonschema/reference.go
tkuhlman: I'm not quite following, can't you just make this
`
nodeMap, ok := node.(map[string]interface{})
if ! ok {
  return nil //or whatever is right}
for key, val := range nodeMap {
`

ikenticus: That syntax creates this error: `unexpected if at end of statement` --- so I fixed it up by adding the necessary semi-colons: `nodeMap, ok := node.(map[string]interface{}); if ! ok { return nil //or whatever is right}; for key, val := range nodeMap {` but then my Atom will auto-format and remove the semicolons and put them on separate lines

tkuhlman: my bad I forgot to tripple backtic the comment, separate lines is what I intended.

PR #22, opened 2018-06-26
Fix error transforming data with nested arrays

tkuhlman:


tkuhlman:
This transformation code is too complicated, I intend to simplify it in a future PR.

PR #23, opened 2018-07-11
Fix handling for a missed error

tkuhlman:


generate/generate.go
```diff
@@ -86,6 +86,9 @@ func buildStructFile(schemaPath, childPath, name, packageName string, embeds []s
 	}
 
 	generated, err := newGeneratedStruct(schema, name, packageName, embeds)
+	if err != nil {
```
enigmatic00: is it possible to include a test as well?

tkuhlman: Not by adding to an existing test. This is the top level function that writes out a file at the end, all of its components are tested but the top level would best be tested by writing a black box test that ran the command through go generate. That is a new set of tests separate from any that currently exist.

PR #26, opened 2018-07-19
Properly walk deep arrays

tkuhlman:


transform/utils.go
```diff
@@ -160,3 +161,10 @@ func schemaDefault(schema json.RawMessage) (interface{}, error) {
 
 	return nil, nil
 }
+
+// replaceIndex takes a path which may include array index values like `a[0].b.c[23].d` with the index values replaced
+// with "*", ie `a[*].b.c[*].d`
+func replaceIndex(path string) string {
```
ahornerr: 👍 Good comment as well

transform/utils_test.go
```diff
@@ -341,3 +341,53 @@ func TestSchemaDefault(t *testing.T) {
...
+			want:        "a[*].b[*]",
+		},
+		{
+			description: "Complicated",
+			path:        "a[3].ab.b[4].c.d[8]",
+			want:        "a[*].ab.b[*].c.d[*]",
+		},
+		{
+			description: "Large Index",
+			path:        "a[13].b[4454]",
```
tkuhlman: For anyone curious it was this case that was failing previously

PR #28, opened 2018-09-10
PENG-6044: Add "-msgp" flag to jstransform for MessagePack support

gannett-dcorzine:


paas-api-pr-manager:

* [PENG-6044](https://jira.gannett.com/browse/PENG-6044): Add msgPack code generation to content-schema
```
With the move to datastore we are using message pack serialization for some of our data. See PENG-4901 for performance numbers on what this brings us. For message pack we are using the library from [https://github.com/tinylib/msgp]

AC:
 * Determine the best approach, either adding message pack code generation to jstransform so it happens as an option in that process or just adding it as an explicit second step within the content-schema repo. For the former consider an optional flag on the jstransform generation.
 * Implement and document the process for getting message pack code generated for new data types.
```

tkuhlman:
Hmm, I wonder why travis didn't update the PR? Something to investigate, at any rate it failed https://travis-ci.org/GannettDigital/jstransform/builds/426842171

generate/generate.go
```diff
@@ -59,44 +60,67 @@ func BuildStructs(schemaPath string, outputDir string) error {
...
+func buildFiles(schemaPath, childPath, name, packageName string, embeds []string, outputDir string, useMessagePack bool) error {
+	outPath, err := buildStructFile(schemaPath, childPath, name, packageName, embeds, outputDir)
+	if err != nil {
+		return fmt.Errorf("failed to build struct file for %q: %v", name, err)
+	}
+
+	if useMessagePack {
+		msgp := "msgp"
+
+		if gopath := os.Getenv("GOPATH"); gopath != "" {
```
tkuhlman: This assumes that msgp is in the GOPATH but I don't think that is a good assumption. Rather than requiring it be installed everywhere I would consider explicitly adding it to Gopkg.toml and then calling it via `go run` in the vendor directory, then fall back to looking for it in the standard $PATH. That means you don't have to install anything to use this when you have the code checked out and automatically get the right version. That is how we often use the package. If you have built jstransform and distributed as a binary you will need msgp as well, the failure to find the executable should be called out as a specific error so the user knows they must install it.

generate/generate.go
```diff
@@ -59,44 +60,67 @@ func BuildStructs(schemaPath string, outputDir string) error {
...
+			msgp = filepath.Join(gopath, "bin", msgp)
+		}
+
+		if err = exec.Command(msgp, "-file", outPath).Run(); err != nil {
```
tkuhlman: Marcelo discovered today that running msgp on a single file does not work if there are structs spread across multiple files. That means you will have to run it on the directory once after all the files are generated.

generate/generate.go
```diff
@@ -59,44 +60,67 @@ func BuildStructs(schemaPath string, outputDir string) error {
...
+		if err := buildFiles(schemaPath, oneOfPath, name, packageName, embeds, outputDir, useMessagePack); err != nil {
+			return err
 		}
 	}
 
 	return nil
 }
 
-func buildStructFile(schemaPath, childPath, name, packageName string, embeds []string, outputDir string) error {
+// Build specified struct file and (if enabled) MessagePack serialization methods.
```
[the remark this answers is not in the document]
gannett-dcorzine: Noted.  I did that with the rewritten code.

generate/generate.go
```diff
@@ -59,44 +60,67 @@ func BuildStructs(schemaPath string, outputDir string) error {
...
+func buildFiles(schemaPath, childPath, name, packageName string, embeds []string, outputDir string, useMessagePack bool) error {
+	outPath, err := buildStructFile(schemaPath, childPath, name, packageName, embeds, outputDir)
+	if err != nil {
+		return fmt.Errorf("failed to build struct file for %q: %v", name, err)
+	}
+
+	if useMessagePack {
+		msgp := "msgp"
```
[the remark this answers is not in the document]
gannett-dcorzine: Good idea, but this is moot after the rewrite.

generate/generate.go
```diff
@@ -13,9 +13,12 @@ import (
 	"strings"
 
 	"github.com/GannettDigital/jstransform/jsonschema"
+	"github.com/tinylib/msgp/gen"
+	"github.com/tinylib/msgp/parse"
+	"github.com/tinylib/msgp/printer"
 )
 
-const disclaimer = "// This file was autogenerated by https://github.com/GannettDigital/jstransform\n\n"
+const disclaimer = "// Code generated by github.com/GannettDigital/jstransform; DO NOT EDIT."
```
enigmatic00: aren't the newlines necessary?

tkuhlman: Deven took them out of the string and put them into the format command where they are used.

generate/generate.go
```diff
@@ -72,9 +75,31 @@ func BuildStructs(schemaPath string, outputDir string) error {
 		}
 	}
 
+	if useMessagePack {
+		mode := (gen.Encode | gen.Decode | gen.Marshal | gen.Unmarshal | gen.Size | gen.Test)
```
tkuhlman: nit: make it a const

tkuhlman:
Also a new unit test like you mentioned should be added before the ticket is done.

generate/generate.go
gannett-dcorzine: Yes, Marcelo and I encountered this problem on Friday, but we didn't know how to resolve it, so I suggested he ask for your input.  It didn't occur to us to try using a directory as the `-file` argument, since the name of the flag implied that it would only work on a single file.  Running msgp on the entire directory at once fixes the problem with assets and draft assets, and that's what I've done with the rewrite.

generate/generate.go
gannett-dcorzine: Yes, since I was changing the string anyhow (to match current Go standards for generated code), I thought it made more sense to have the newlines in the format string instead of this constant string.

gannett-dcorzine:
Good point.  I've created a revert PR for the Docker image, since it's no longer needed.

https://github.com/GannettDigital/api-services-common-ci/pull/41

generate/generate.go
gannett-dcorzine: Will do.

generate/test_data/test_data_msgp.go.out
```diff
@@ -0,0 +1,755 @@
+package test_data
+
+// Code generated by github.com/tinylib/msgp DO NOT EDIT.
+
+import (
+	"github.com/tinylib/msgp/msgp"
```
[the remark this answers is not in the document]
gannett-dcorzine: No, because that would create an unnecessary dependency on a fork that we hope will be temporary.  Using our version would be appropriate if we decide to make the fork permanent, but otherwise it's better not to -- we can override the source of the upstream package to point to our fork in projects that use the generated MessagePack code.

PR #33, opened 2018-10-15
PENG-7134 return error when dates fail to convert

tomtoday:


paas-api-pr-manager:

* [PENG-7134](https://jira.gannett.com/browse/PENG-7134): Anything but strict RFC3339 dates should produce errors on transform
```
As a developer, jstransform should propagate an error if it cannot transform a string into a date-time based on the RFC3339 format.

As an example: this JSON is almost correct:

{code:javascript}
{
     "id": "6633dcec-b521-45f2-9514-30b3f622c79a",
     "date": "1977-06-24",
     "parentId": "b8175649-d3a0-435e-8254-140dfc0b453e",
     "name": "business",
     "displayName": "Business",
     "path": "money/business",
     "vocabulary": "UsaToday",
     "attributes": {
          "awsPath": [
               "money/business"
          ],
          "awsPathFront": [
               "money/business/main"
          ],
          "modified": [
               "1977-06-24"
          ],
          "created": [
               "2012-06-07T00:08:21.111-04:00"
          ]
     }
}
{code}

The string "1977-06-24" can be parsed into a date-time but with jstransform's strict RFC3339 formatting it will error when unmarshalled. The error in this case is silent and a zero value for the time is returned. The error should bubble up.
```

transform/instance_test.go
```diff
@@ -496,7 +512,15 @@ func TestScalarTransform(t *testing.T) {
 		}
 
 		got, err := st.transform(test.in)
+
 		if err != nil {
+			if err.Error() == test.wantError {
+				break // pass
```
tkuhlman: continue rather then break. It happens to be the last test right now but in the future more could be added later.

PR #34, opened 2018-10-18
Test for and handle transform concurrency

tkuhlman:


transform/transformer_test.go
```diff
@@ -379,23 +380,31 @@ func TestSaveValue(t *testing.T) {
...
-		case !test.wantErr && err != nil:
-			t.Errorf("Test %q - got error, want nil: %v", test.description, err)
-		case !reflect.DeepEqual(got, test.want):
-			t.Errorf("Test %q - got\n%s\nwant\n%s", test.description, got, test.want)
+				switch {
+				case wantErr && err != nil:
+					return
+				case wantErr && err == nil:
+					t.Errorf("Test %q - got nil, want error", description)
+				case !test.wantErr && err != nil:
```
abaehre: should this also be wantErr and not test.wantErr?

tkuhlman: good catch, updating

PR #38, opened 2018-11-28
PENG-7770 - Avoid crashing when 'nil' value is in a JSON array of schema.

gannett-ggreer:


paas-api-pr-manager:

* [PENG-7770](https://jira.gannett.com/browse/PENG-7770): Tag schema properties mismatch
```
The list of properties allowed on a tag:
`/properties/attributes/properties/*`

is not the same as the list of properties allowed on a referenced tag:
`/properties/relationships/items/properties/referencedEntity/properties/attributes/properties/*`

but they should be the same list.
```

PR #39, opened 2018-12-03
PENG-7977 - sort errors returned to enable accurate testing

abaehre:


paas-api-pr-manager:

* [PENG-7977](https://jira.gannett.com/browse/PENG-7977): Update content-schema to use the shared schema tests
```
As we have split the schema repos into data group focused groups we have been working on a common set of schema test tools. These are found in content-pipeline/schema

AC:
 * Move to using the shared schema testing tools for content-schema
 * Improve the shared code as needed while maintaining clarity.
```

jsonschema/validator_test.go
```diff
@@ -50,8 +51,11 @@ func TestValidate(t *testing.T) {
 		}
 
 		got, err := v.Validate(raw)
-
-		if (err != nil) != test.wantErr {
+		if err != nil && test.wantErr != nil {
+			if err.Error() != test.wantErr.Error() {
+				t.Errorf("Test %q - got error %v, want error %t", test.description, err, test.wantErr)
```
tomtoday: Minor nit: since you are comparing the error strings, you could output the error string in the failure message to improve readability of the test result. Your choice.

abaehre: since we are using Errorf I'm pretty sure it'll be equivalent.

abaehre:
@tkuhlman @tomtoday - do we want this to be a minor or patch bump? This would technically break tests and I don't necessarily want people to accidentally get stuck updating tests because they dep ensured.

jsonschema/validator_test.go
tomtoday: I tried it out, it was a bit difficult to read:
```
=== RUN   TestValidate
--- FAIL: TestValidate (0.01s)
    validator_test.go:56: Test "invalid image JSON" - got error invalid schema: [URL: URL is required caption: caption is required credit: credit is required crops: crops is required datePhotoTaken: datePhotoTaken is required orientation: orientation is required originalSize: originalSize is required type: type must be one of the following: "image"], want error &{%!t(string=Invalid schema: [URL: URL is required caption: caption is required credit: credit is required crops: crops is required datePhotoTaken: datePhotoTaken is required orientation: orientation is required originalSize: originalSize is required type: type must be one of the following: "image"])}
```
not a huge deal.

tomtoday: basically `got error invalid schema: ...` vs `want error &{%!t(string=Invalid schema: ...`

tkuhlman:
Either minor or patch bump is fine by me. I don't think the order of the string output of errors is something we ever have put any assurances around so it isn't really a breaking change in mind.

PR #40, opened 2018-12-04
PENG-7977 - add an interface to be implemented and implement it for t…

abaehre:
…ransformer

I think the names could probably be better. Open to ideas.

### DIFF
--- a/transform/transformer.go
+++ b/transform/transformer.go
@@ -14,7 +14,7 @@ import (
 	"github.com/buger/jsonparser"
 )
 
-// JSONTransformer - a type implemented by the jstransform.Transformer and the transformers in content-schema
+// JSONTransformer - a type implemented by the jstransform.Transformer
 type JSONTransformer interface {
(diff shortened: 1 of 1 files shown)


paas-api-pr-manager:

* [PENG-7977](https://jira.gannett.com/browse/PENG-7977): Update content-schema to use the shared schema tests
```
As we have split the schema repos into data group focused groups we have been working on a common set of schema test tools. These are found in content-pipeline/schema

AC:
 * Move to using the shared schema testing tools for content-schema
 * Improve the shared code as needed while maintaining clarity.
```

transform/transformer.go
```diff
@@ -14,6 +14,11 @@ import (
 	"github.com/buger/jsonparser"
 )
 
+// JSONTransformer - a type implemented by the jstransform.Transformer and the transformers in content-schema
```
tkuhlman: nit: content-schema is a private repo, this is a public repo so for many people that comment won't quite make sense.

PR #41, opened 2018-12-11
PENG-8142 - add a debug log when there is an error

abaehre:


paas-api-pr-manager:

* [PENG-8142](https://jira.gannett.com/browse/PENG-8142): Investigate issue in jstransform reading input fields with special chars in the name
```
As an API Services engineer, I want to figure out what is causing an issue in jstransform where if the fromInput.from field contains special characters, it fails silently. This manifested itself when attempting to transform the following fields for frontAttribute and moduleAttribute additions to fronts:

front-list-module-position to frontListModulePosition
 og:image to ogImage

Acceptance:
 1.) investigate what appears to be an issue in jstransform when a source field contains a special character
 2.) fix issue if able, otherwise create additional tickets to correct as needed
```

transform/transform.go
```diff
@@ -89,6 +92,7 @@ func (ti *transformInstruction) transform(in interface{}, fieldType string, modi
 	}
 	rawValue, err := jsonpath.Get(path, in)
 	if err != nil {
+		level.Debug(log.NewLogfmtLogger(os.Stderr)).Log("Error in JSONPath during transform: %v", err)
```
ahornerr: Perhaps we could initialize the logger elsewhere? Not sure how often transform errors occur but could cause a performance penalty if it happens frequently.

abaehre: Maybe? This only happens when the jsonpath lib fails. That only should fail when it can't find the field or the gval library fails. So only when we have fields done incorrectly.

PR #42, opened 2018-12-12
PENG-8142 - were ignoring a valid jsonpath option in transform instru…

abaehre:
…ctions

### DIFF
--- a/transform/transformer_test.go
+++ b/transform/transformer_test.go
@@ -26,204 +26,204 @@ var (
 		want                json.RawMessage
 		wantErr             bool
 	}{
-		// {
-		// 	description:         "Use basic transforms, copy from input and default to build result",
-		// 	schema:              imageSchema,
-		// 	transformIdentifier: "cumulo",
-		// 	in: json.RawMessage(`
-		// 					{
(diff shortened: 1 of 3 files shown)


paas-api-pr-manager:

* [PENG-8142](https://jira.gannett.com/browse/PENG-8142): Investigate issue in jstransform reading input fields with special chars in the name
```
As an API Services engineer, I want to figure out what is causing an issue in jstransform where if the fromInput.from field contains special characters, it fails silently. This manifested itself when attempting to transform the following fields for frontAttribute and moduleAttribute additions to fronts:

front-list-module-position to frontListModulePosition
 og:image to ogImage

Acceptance:
 1.) investigate what appears to be an issue in jstransform when a source field contains a special character
 2.) fix issue if able, otherwise create additional tickets to correct as needed
```

transform/transformer_test.go
```diff
@@ -25,203 +26,219 @@ var (
 		want                json.RawMessage
 		wantErr             bool
 	}{
+		// {
```
tkuhlman: uncomment the existing tests

PR #43, opened 2018-12-12
remove logging

abaehre:


transform/transform.go
```diff
@@ -92,7 +89,6 @@ func (ti *transformInstruction) transform(in interface{}, fieldType string, modi
 	}
 	rawValue, err := jsonpath.Get(path, in)
 	if err != nil {
-		level.Debug(log.NewLogfmtLogger(os.Stderr)).Log("Error in JSONPath during transform: %v", err)
```
tkuhlman: Note line 83, explicitly mentions this is not an error condition.

PR #46, opened 2019-02-22
Allow renaming of generated structure names.

gannett-ggreer:


generate/generate.go
```diff
@@ -35,7 +35,7 @@ const msgpMode = gen.Encode | gen.Decode | gen.Marshal | gen.Unmarshal | gen.Siz
 // The package name is set to the outputPath directory name.
 //
 // NOTE: If oneOf/allOf entries exist than any JSON schema instances in the root schema file will be skipped.
-func BuildStructs(schemaPath string, outputDir string, useMessagePack bool) error {
+func BuildStructs(schemaPath string, outputDir string, useMessagePack bool, nameMap map[string]string) error {
```
enigmatic00: this should bump hte minor/major version or a backwards compatible method signature also needs to be implemented

gannett-ggreer: It is a good question what the expected API of this is, given it is packaged as a program.  Since this function is capitalized it could have potentially been used so I'll make it a wrapper function.

gannett-ggreer:
Oh so that's what the little "recycle" button does...

generate/generate.go
```diff
@@ -34,8 +39,10 @@ const msgpMode = gen.Encode | gen.Decode | gen.Marshal | gen.Unmarshal | gen.Siz
 //
 // The package name is set to the outputPath directory name.
 //
+// The nameMap allows specifying the type name of the structure for each JSON file.
+//
 // NOTE: If oneOf/allOf entries exist than any JSON schema instances in the root schema file will be skipped.
-func BuildStructs(schemaPath string, outputDir string, useMessagePack bool) error {
+func BuildStructsRename(schemaPath string, outputDir string, useMessagePack bool, nameMap map[string]string) error {
```
enigmatic00: the shim is supplying a nil for nameMap. Won't line 72 error out with a nilpointer?

gannett-ggreer: The tests pass a `nil` pointer for two of the cases, so it is fine.  If you paste this into a Go Playground it won't crash:
```
package main

import "fmt"

func main() {
	var names map[string]string
	if got, ok := names["foo"]; ok {
		fmt.Println("ok: " + got)
	} else {
		fmt.Println("no match")
	}	
}
```

PR #47, opened 2019-03-14
APIS-2165 Add option to concatenate with delimiter

adamhostettler:


paas-api-pr-manager:

* [APIS-2165](https://jira.gannett.com/browse/APIS-2165): SSTS path different from attributes
```
Here you can see that a section exists, but path is empty: 

[https://content-api.gannettdigital.com/graphiql?siteCode=USAT&type=asset&query=query%20Gallery%28%24id%3A%20String%21%29%20%7B%20asset%28id%3A%20%24id%29%20%7B%20id%20ssts%20%7B%20leafName%20storysubject%20section%20path%20subsection%20subtopic%20taxonomyEntityDisplayName%20topic%20%7D%20%7D%20%7D%20&variables=%7B%20%22id%22%3A%20%221410095%22%20%7D&operationName=Gallery]

web call: 

[http://web-api-west.production.gannettdigital.com/v4/assets/1410095?consumer=testing&sitereferrerid=1&global=true]

Looks like path is being pulled from somewhere: [https://github.com/GannettDigital/content-api-asset/blob/master/schema/output/assets/base.json#L653]

We can concat these attributes ourselves, but would like to know why path is null here. 

 

AC

new operator in jstransform to join with a delimiter

content-api-asset updated to join for ssts path

content-ingestor-api updated
```

transform/transform.go
```diff
@@ -116,13 +116,19 @@ func (ti *transformInstruction) transform(in interface{}, fieldType string, modi
...
 type transformInstructionsJSON struct {
-	From   []*transformInstruction `json:"from"`
-	Method string                  `json:"method"`
+	From          []*transformInstruction `json:"from"`
+	Method        string                  `json:"method"`
+	MethodOptions methodOptions           `json:"methodOptions"`
+}
+
+type methodOptions struct {
+	ConcatenateDelimiter interface{} `json:"concatenateDelimiter"`
```
tkuhlman: That should be a string not an empty interface, it makes the code nicer. Also though concat can theoretically work on slices or other types I think it is fine to say that the delimiter is only supported for strings.

transform/utils.go
```diff
@@ -15,7 +15,7 @@ import (
 var indexRe = regexp.MustCompile(`\[([\d]+)\]`)
 
 // Concat will combine any two arbitrary values, though only strings are supported for non-trivial concatenation.
-func concat(a, b interface{}) (interface{}, error) {
+func concat(a, b, delimiter interface{}) (interface{}, error) {
```
tkuhlman: I think that the delimiter should be a string. I think the reason that a and b are empty interface is that there type is not known but that isn't true of the delimiter.

transform/transform_test.go
```diff
@@ -177,6 +177,27 @@ func TestTransformInstructions(t *testing.T) {
 			in:   testRaw,
 			want: "outout2",
 		},
+		{
```
tkuhlman: Nice to add it here, I think it is also sufficiently different that you should add a test case to TestTransformUnmarshal

adamhostettler:
@tkuhlman I think i've addressed your comments so far, ready for another look when you get a chance

PR #49, opened 2019-04-13
Allow field renames

tomtoday:
Valid JSON property names are not always valid Go identifiers. A common example is a property name that starts with a number. This cannot be turned into a field in the Go struct generated by this tool.
With this change you can supply key value pairs to rename JSON property names to something that is a valid Go identifier. For example if you had a property name in your JSONschema file called `"16_9"` you could call go generate like this:
```
//go:generate jstransform -renameFields 16_9=SixteenByNine /path/to/the/schema.json
```


generate/generate.go
```diff
@@ -310,22 +322,27 @@ func (gen *generatedStruct) write(w io.Writer) error {
...
 		f := &extractedField{jsonName: tree[0], jsonType: "object", name: exportedName(tree[0]), fields: make(map[string]*extractedField)}
 		fields[tree[0]] = f
-		if err := addField(f.fields, tree[1:], inst); err != nil {
+		if err := addField(f.fields, tree[1:], inst, fieldRenameMap); err != nil {
 			return fmt.Errorf("failed field %q: %v", tree[0], err)
 		}
 		return nil
 	}
 
+	fieldName := fieldRenameMap[tree[0]]
```
abaehre: is there an instance where tree will have a length of 0?

tomtoday: Good question, I'll see if I can demonstrate that.

abaehre: Thanks Tom

generate/generate.go
```diff
@@ -41,8 +46,12 @@ func BuildStructs(schemaPath string, outputDir string, useMessagePack bool) erro
 //
 // The nameMap allows specifying the type name of the structure for each JSON file.
 //
+// The fieldNameMap is used to provide alternate names for fields in the resulting structs.
+// The property names in the JSON tags for these structs remains the same as supplied.
+// This can be used to accommodate names that are valid JSON but not valid Go identifiers
+//
 // NOTE: If oneOf/allOf entries exist than any JSON schema instances in the root schema file will be skipped.
-func BuildStructsRename(schemaPath string, outputDir string, useMessagePack bool, nameMap map[string]string) error {
+func BuildStructsRenameFields(schemaPath string, outputDir string, useMessagePack bool, nameMap map[string]string, fieldNameMap map[string]string) error {
```
tkuhlman: This function now has 5 arguments. Have we reached the point where we should have struct for the arguments rather than continuing to make the function signature larger?

main.go
```diff
@@ -35,17 +37,20 @@ func (mf *mapFlags) Set(value string) error {
...
+	flag.Var(&renameFields, "renameFields", "Override generated name of structure; use '-renameFields old=new'.")
 	flag.BoolVar(&useMessagePack, "msgp", false, "generate MessagePack serialization methods")
 
 	flag.Parse()
 
 	args := flag.Args()
 
 	if len(args) < 1 {
-		fmt.Printf("Usage: %s [-msgp] <JSON Schema Path> [output directory]\n", path.Base(os.Args[0]))
+		fmt.Printf("Usage: %s [-msgp] [-rename[ [-renameFields] <JSON Schema Path> [output directory]\n", path.Base(os.Args[0]))
```
tkuhlman: `[-rename[` should be `[-rename]` maybe even `[-rename k=v]` to indicate that a value must follow it. Also FWIW the pflag package here supports flag.StringArrayVar which may be more appropriate here to allow multiple options.

tomtoday: Thanks, I can fix that

generate/generate.go
tomtoday: I'll see what I can do

main.go
tomtoday: adding pflag (https://github.com/spf13/pflag) can be done in another PR, later.

PR #50, opened 2019-05-06
APIS-2601 add xml transform functionality 

jstorer:


paas-api-pr-manager:

* [APIS-2601](https://jira.gannett.com/browse/APIS-2601): Add POC for XPath Sports-Aggregator parsing to JSTransform
```
Add the Proof of concepts from APIS-2600 and APIS-2538 to the jstransform library.

Look into how the code can interface with already existing jstransform functionality.

AC - 
POCs have been added to jstransform
```

transform/transformer.go
```diff
@@ -114,6 +182,29 @@ func (tr *Transformer) findParent(path string) (instanceTransformer, error) {
 	return parent, nil
 }
 
+// findParent walks the instanceTransformer tree to find the parent of the given path
+func (tr *TransformerXML) findParent(path string) (instanceTransformer, error) {
```
ahornerr: Very interesting. I hope this works well.

transform/utils.go
```diff
@@ -84,6 +85,9 @@ func convertBoolean(raw interface{}) (interface{}, error) {
 		return t > 0, nil
 	case nil:
 		return nil, nil
+	case []*xmlquery.Node:
+		node := raw.([]*xmlquery.Node)
```
ahornerr: Don't need to type assert here if you do the switch/case trick

transform/utils.go
```diff
@@ -109,6 +113,15 @@ func convertNumber(raw interface{}) (interface{}, error) {
 		return nil, fmt.Errorf("failed to convert string %q to number", t)
 	case int, float32, float64:
 		return raw, nil
+	case []*xmlquery.Node:
+		node := raw.([]*xmlquery.Node)
```
ahornerr: Same as above

transform/utils.go
```diff
@@ -125,6 +138,9 @@ func convertDateTime(raw interface{}) (interface{}, error) {
 		return time.Unix(int64(t), 0).UTC(), nil
 	case float64:
 		return time.Unix(int64(t), 0).UTC(), nil
+	case []*xmlquery.Node:
+		node := raw.([]*xmlquery.Node)
```
ahornerr: Same as above

transform/utils.go
```diff
@@ -145,6 +161,9 @@ func convertString(raw interface{}) (interface{}, error) {
 		return strconv.FormatFloat(t64, 'f', -1, 32), nil
 	case float64:
 		return strconv.FormatFloat(t, 'f', -1, 64), nil
+	case []*xmlquery.Node:
+		node := raw.([]*xmlquery.Node)
```
ahornerr: Same as above

transform/operations.go
```diff
@@ -204,12 +205,18 @@ func (s *split) init(args map[string]string) error {
...
-	in, ok := raw.(string)
-	if !ok {
-		return nil, errors.New("split only supports strings")
+	var stringToSplit string
+
+	switch t := raw.(type) {
+	case []*xmlquery.Node:
+		stringToSplit = t[0].InnerText()
+	case string:
+		stringToSplit = raw.(string)
```
ahornerr: You can use `t` here instead of type asserting to a string

transform/transformer.go
```diff
@@ -114,6 +189,29 @@ func (tr *Transformer) findParent(path string) (instanceTransformer, error) {
...
```
tkuhlman: This looks identical to the one from the JSON transformer. Unless I'm missing something looks like you could make a generic function that passes in tr.root and returns a parent.

transform/instance.go
```diff
@@ -332,12 +367,15 @@ func (st *scalarTransformer) transform(in interface{}, modifier pathModifier) (i
...
-	if err == nil {
-		newValue, err := convert(rawValue, st.jsonType)
-		// if there is a conversion error fall through to the default
-		if newValue != nil {
-			return newValue, err
+	//an xmlNode will never have a JSONPath value
+	if _, isXmlNode := in.(*xmlquery.Node); !isXmlNode {
+		rawValue, err := jsonpath.Get(path, in)
+		if err == nil {
+			newValue, err := convert(rawValue, st.jsonType)
```
abaehre: do we want to do anything with this error?

abaehre: and we can probs combine lines here.

transform/transformer.go
```diff
@@ -146,6 +244,38 @@ func (tr *Transformer) walker(path string, value json.RawMessage) error {
 	return nil
 }
 
+// walker is a WalkFunc for the Transformer which builds an representation of the fields and transforms in the schema.
+// This is later used to do the actual transform for incoming data
+func (tr *TransformerXML) walker(path string, value json.RawMessage) error {
```
tkuhlman: Also looking nearly identical, maybe the thing to do is to combine the transformers adding a field that says if the input is XML or JSON and then just pass that down to the new*Transformer functions.

transform/operations.go
```diff
@@ -204,12 +205,18 @@ func (s *split) init(args map[string]string) error {
...
 
 func (s *split) transform(raw interface{}) (interface{}, error) {
-	in, ok := raw.(string)
-	if !ok {
-		return nil, errors.New("split only supports strings")
+	var stringToSplit string
+
+	switch t := raw.(type) {
+	case []*xmlquery.Node:
+		stringToSplit = t[0].InnerText()
```
abaehre: length check?

transform/transformer_test.go
```diff
@@ -426,6 +427,100 @@ func TestTransformer(t *testing.T) {
 	}
 }
 
+func TestNewTransformerXML(t *testing.T) {
```
tkuhlman: Yah Tests!

transform/utils.go
```diff
@@ -84,6 +85,8 @@ func convertBoolean(raw interface{}) (interface{}, error) {
 		return t > 0, nil
 	case nil:
 		return nil, nil
+	case []*xmlquery.Node:
+		return strconv.ParseBool(t[0].InnerText())
```
tkuhlman: I think this should happen before calling the utils, ie the `t[0].InnerText()` part. At the least if it stays this way you will need to expand the unit tests for conversion.

transform/transform.go
```diff
@@ -32,11 +33,13 @@ type transformOperationJSON struct {
 type transformInstruction struct {
 	// For JSONPath format see http://goessner.net/articles/JsonPath/
 	jsonPath   string
+	xmlPath    string
 	Operations []transformOperation `json:"operations"`
 }
 
 type transformInstructionJSON struct {
 	JsonPath   string                   `json:"jsonPath"`
+	XmlPath    string                   `json:"xmlPath"`
```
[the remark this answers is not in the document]
abaehre: and JSONPath?

transform/transformer.go
ahornerr: I like this idea

transform/transform.go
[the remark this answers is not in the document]
tkuhlman: oh yeah that should have been JSONPath, don't look at the git blame on that one.

transform/instance.go
```diff
@@ -32,18 +33,19 @@ type instanceTransformer interface {
...
+// in both cases the output will be JSON
 type arrayTransformer struct {
 	childTransformer instanceTransformer
 	defaultValue     []interface{}
 	jsonPath         string
 	transforms       *transformInstructions
 }
 
-func newArrayTransformer(path, transformIdentifier string, raw json.RawMessage) (*arrayTransformer, error) {
+func newArrayTransformer(path, transformIdentifier string, raw json.RawMessage, _ string) (*arrayTransformer, error) {
```
tkuhlman: Hmm, I was thinking if you set the type into the arrayTransformer struct here you don't have to pass it in for other functions just in the new part.

transform/instance.go
```diff
@@ -344,3 +444,39 @@ func (st *scalarTransformer) transform(in interface{}, modifier pathModifier) (i
...
+		}
+	}
+
+	// 2. Fall back to the JSON Schema default value.
+	return st.defaultValue, nil
+}
+
+// transform routes to the correct scalar transform type
+func (st *scalarTransformer) transform(in interface{}, modifier pathModifier, transformType string) (interface{}, error) {
+	if transformType == "json" {
```
tkuhlman: This is what I was thinking but you had to change transform command to get this to work by adding another argument. I thought it may be simpler to just set the transform type in the newScalarTransform function. Then here it is just like this but `if st.transformType == "json"`

transform/transformer.go
```diff
@@ -27,19 +38,20 @@ type Transformer struct {
...
+	format              inputFormat
 }
 
 // NewTransformer returns a Transformer using the schema given.
 // The transformIdentifier is used to select the appropriate transform section from the schema.
-func NewTransformer(schema *jsonschema.Schema, tranformIdentifier string) (*Transformer, error) {
-	tr := &Transformer{schema: schema, transformIdentifier: tranformIdentifier}
-
+// The format is used to execute the correct transforms ('JSON' or 'XML')
+func NewTransformer(schema *jsonschema.Schema, tranformIdentifier string, format inputFormat) (*Transformer, error) {
```
tkuhlman: This is an exported function and we don't want to do any breaking changes. Basically keep this as is but assume the format is JSON and add a NewXMLTransformer which sets the format to xml. That avoids it being a breaking change.

jstorer: oh makes sense, incoming commit

transform/transform.go
```diff
@@ -27,16 +29,18 @@ type transformOperationJSON struct {
 	Args map[string]string `json:"args"`
 }
 
-// transformInstruction defines a JSONPath for a transform and an optional set of operations to be performed on the
+// transformInstruction defines a JSONPath and XMLPath for a transform and an optional set of operations to be performed on the
 // data from that path.
 type transformInstruction struct {
 	// For JSONPath format see http://goessner.net/articles/JsonPath/
-	jsonPath   string
+	JSONPath   string
```
ahornerr: Looks like these could both be combined into just `path` since you're passing around the format to the transformer

transform/transform.go
```diff
@@ -78,12 +83,53 @@ func (ti *transformInstruction) UnmarshalJSON(data []byte) error {
...
+		return nil, nil
+	}
+	rawValue := xmlNode
+
+	var (
+		value interface{}
+		err   error
+	)
+
+	if len(xmlNode) == 1 {
```
ahornerr: 👍

transform/transform.go
```diff
@@ -78,12 +83,53 @@ func (ti *transformInstruction) UnmarshalJSON(data []byte) error {
...
+		value, err = convert(xmlNode[0].InnerText(), fieldType)
+	} else {
+		value, err = convert(xmlNode, fieldType)
+	}
+
+	if err != nil {
+		// In some cases the conversion is helpful but in others like before a max operation it isn't
+		value = rawValue
```
ahornerr: This seems a little weird to me but i'll allow it.

ahornerr: I guess it makes sense since it's a fallback if the convert method errors

transform/transform.go
```diff
@@ -113,6 +159,20 @@ func (ti *transformInstruction) transform(in interface{}, fieldType string, modi
 	return value, nil
 }
 
+// transform runs the instructions in this object returning the new transformed value or an error if unable to.
+// It handles the logic for finding the value to be transformed and chaining the Operations.
+// It will not error if the value is not found, rather it returns nil for the value.
+// If a conversion or operation fails an error is returned.
+func (ti *transformInstruction) transform(in interface{}, fieldType string, modifier pathModifier, format inputFormat) (interface{}, error) {
```
ahornerr: 👍

transform/transformer.go
```diff
@@ -3,17 +3,28 @@
...
 
+	"github.com/antchfx/xmlquery"
+
 	"github.com/GannettDigital/jstransform/jsonschema"
 
 	"github.com/buger/jsonparser"
 )
 
+// inputFormat denotes the type of transform to perfrom, the options are 'JSON' or 'XML'
+type inputFormat string
```
ahornerr: Cool use of type aliases

transform/transform.go
abaehre: also I think I agree with Anderson's comment

transform/transform.go
```diff
@@ -78,12 +84,53 @@ func (ti *transformInstruction) UnmarshalJSON(data []byte) error {
...
+	node, ok := in.(*xmlquery.Node)
+	if !ok {
+		return nil, errors.New("Error converting input to *xmlquery.Node")
+	}
+
+	xmlNode := xmlquery.Find(node, path)
+	if xmlNode == nil {
+		return nil, nil
+	}
+	rawValue := xmlNode
```
abaehre: actually. Followup question is if that can ever return nil - and I actually see no.

transform/transform.go
```diff
@@ -78,12 +84,53 @@ func (ti *transformInstruction) UnmarshalJSON(data []byte) error {
...
+	if modifier != nil {
+		path = modifier(path)
+	}
+
+	node, ok := in.(*xmlquery.Node)
+	if !ok {
+		return nil, errors.New("Error converting input to *xmlquery.Node")
+	}
+
+	xmlNode := xmlquery.Find(node, path)
```
abaehre: https://github.com/antchfx/xmlquery/blob/master/query.go#L73 <- this worries me. Would it be worth it to add a recover?

transform/transform.go
abaehre: also - we went from private to public. Which is probably also wrong.

transform/transformer.go
```diff
@@ -91,6 +124,33 @@ func (tr *Transformer) Transform(raw json.RawMessage) (json.RawMessage, error) {
 	return out, nil
 }
 
+func (tr *Transformer) xmlTransform(raw []byte) ([]byte, error) {
```
abaehre: this is actually a json.RawMessage

transform/instance.go
```diff
@@ -107,13 +111,58 @@ func (at *arrayTransformer) baseValue(in interface{}, path string, modifier path
...
+func (at *arrayTransformer) baseValueXML(in interface{}, path string, modifier pathModifier) ([]interface{}, bool, error) {
+	// 1. Use a transform if it exists
+	if at.transforms != nil {
+		rawValue, err := at.transforms.transform(in, "array", modifier, at.format)
+		if err != nil {
+			return nil, false, err
+		}
+
+		//if rawValue is an array of xml nodes we need to append them to newValue for return as []interface{}
+		xmlNodeArray, ok := rawValue.([]*xmlquery.Node)
```
abaehre: is this block to just change the underlying type of the interface to `xmlquery.Node`s?

transform/instance.go
```diff
@@ -107,13 +111,58 @@ func (at *arrayTransformer) baseValue(in interface{}, path string, modifier path
...
+	// 2. Fall back to the JSON Schema default value.
+	if at.defaultValue != nil {
+		return at.defaultValue, true, nil
+	}
+	return nil, false, nil
+}
+
+// baseValue routes to the correct arrayTransformer.baseValue format
+func (at *arrayTransformer) baseValue(in interface{}, path string, modifier pathModifier) ([]interface{}, bool, error) {
+	if at.format == jsonInput {
```
abaehre: I seem to ask this question a lot, but is it worth it to create a map for these? a `map[inputFormat]baseValueFunc`? Hypothetically we only will be doing XML and JSON is just a thought

transform/instance.go
```diff
@@ -163,18 +212,72 @@ func (at *arrayTransformer) transform(in interface{}, modifier pathModifier) (in
...
+	if at.childTransformer == nil {
+		return base, nil
+	}
+
+	oldPath := path + "[*]"
+	newArray := make([]interface{}, 0, len(base))
+
+	for i := range base {
+		currentPath := path + fmt.Sprintf("[%d]", i)
+		var childValue interface{}
```
abaehre: why did this get separated from the line below?

transform/transform.go
abaehre: also the comment for XPath is next to the JSONpath one :P

transform/transform.go
```diff
@@ -78,12 +84,53 @@ func (ti *transformInstruction) UnmarshalJSON(data []byte) error {
...
+		value = rawValue
+	}
+	if value == nil {
+		return nil, nil
+	}
+
+	for _, op := range ti.Operations {
+		value, err = op.transform(value)
+		if err != nil {
+			return nil, fmt.Errorf("failed operation on value from JSONPath %q: %v", path, err)
```
abaehre: from XMLPath?

transform/transform.go
abaehre: do we want to add a nil check here? I see one in the other method to save some time on trying to convert nil

PR #54, opened 2019-05-31
APIS-2801 xmltransform handle single element // paths correctly

jstorer:


paas-api-pr-manager:

* [APIS-2801](https://jira.gannett.com/browse/APIS-2801): update XML transform to handle // items with only one element
```
xmltransform is incorrectly handling //paths that only contain one element, a fix should be made.
```

PR #55, opened 2019-06-03
fix bug for grabbing an array of attributes using xPath contains()

jstorer:


valeramaniuk:
LGTM, but what do I know

PR #56, opened 2019-06-06
Allow XML objects to use refs so that items inside object can use a ref with paths relative to the object node

jstorer:


transform/instance_test.go
```diff
@@ -226,6 +226,7 @@ func TestArrayTransform(t *testing.T) {
 				format:   jsonInput,
 				childTransformer: &objectTransformer{
 					jsonPath: "$.otherCrops[*][*]",
+					format:   jsonInput,
```
ahornerr: Is there a similar test for XML?

transform/instance.go
```diff
@@ -367,6 +367,76 @@ func (ot *objectTransformer) transform(in interface{}, modifier pathModifier) (i
...
+	}
+	var newValue map[string]interface{}
+
+	// For the object use a transform if it exists or the default or an empty map
+	if ot.transforms != nil {
+		rawValue, err := ot.transforms.transform(in, "object", modifier, ot.format)
+		if err != nil {
+			return nil, err
+		}
+		newValue = make(map[string]interface{})
```
ahornerr: I'm not sure this needs to be defined here, the code in lines 400-406 seems to take care of it

jstorer: good call

transform/instance_test.go
jstorer: off handedly its covered in the transformer_test.go , but maybe it would be worth building some explicitly in here as well

transform.adoc
```diff
@@ -37,7 +37,7 @@ The transform object has the following properties:
 
 - The jsonPath selector is an identifier of the location of the _input_ path. All paths should start with the root object `$`. The exception here is for an Array of objects, where the relative `@` operator should be used on the leaf objects to signify the location is relative to the parent transform (more on this below)
 
-- The xPath selector is an identifier of the location of the _input_ path. All paths should start with the root object `/`. The exception here is for an Array of objects, where `//` can be used to select a node and all of its descendants. The xPath inside the array can then be relative to this node.
+- The xPath selector is an identifier of the location of the _input_ path. All paths should start with the root object `/`. The exception here is for an Array of objects, where `//` can be used to select a node and all of its descendants. The xPath inside the array can then be relative to this node. An object can also use an xPath that will result in the selection of a node in which the fields inside the object will be able to be referenced relative the base object's node.
```
tkuhlman: Maybe rather than `All paths should start with the root object` it is time to say something like `Many paths will start with the root object '/' but items in arrays and objects can be relative. . . .`

transform/instance.go
```diff
@@ -317,7 +317,7 @@ func (ot *objectTransformer) selectChild(key string) instanceTransformer { retur
 
 // transform retrieves the value for this object by building the value for the base object and then adding in any
```
tkuhlman: You changed the function name but need to fix up the godocs to match, this is true for all xml and updated transform function as well.

transform/instance.go
```diff
@@ -367,6 +367,75 @@ func (ot *objectTransformer) transform(in interface{}, modifier pathModifier) (i
...
+		case []*xmlquery.Node:
+			if len(v) > 0 {
+				in = v[0]
+			}
+		default:
+			return nil, errors.New("non xml node returned from object transform")
+		}
+	}
+
+	if newValue == nil {
```
tkuhlman: Looks like newValue is always nil since it was just defined up above. I would drop the if statement and consider bring the variable definition here since it isn't used until this point.

transform/instance.go
```diff
@@ -367,6 +367,75 @@ func (ot *objectTransformer) transform(in interface{}, modifier pathModifier) (i
...
+
+	// For the object use a transform if it exists or the default or an empty map
+	if ot.transforms != nil {
+		rawValue, err := ot.transforms.transform(in, "object", modifier, ot.format)
+		if err != nil {
+			return nil, err
+		}
+
+		switch v := rawValue.(type) {
+		case *xmlquery.Node:
```
tkuhlman: In the array transformer it works not by redefining `in` but by making a path modifier that is passed to all the children. The path modifier then only modifies relative paths. If I'm remembering the value of in correctly doesn't changing it out at this point break the possibility non-relative paths on transforms within the object? To state it another way if one of the fields in the object had a transform that should pull from another path then that path won't be found right?

If that is right then I think we need to first ask if that use case is needed and if it isn't document that anytime you specify a transform on an object all child transforms must be relative.

jstorer: i have documented the limitations and will make a ticket to revisit possibly resolving this in the future. As it stands now if an xmlPath is declared on an object all paths inside that object must be relative to the initial declared path, so things above the level of the declared path of the object will be inaccessible inside the object. I believe all SDI use cases will be covered as is.

PR #57, opened 2019-06-10
handle a case in which an xmlTransform uses an object in which the tr…

jstorer:
…ansform path is not found.

### DIFF
--- a/transform/instance.go
+++ b/transform/instance.go
@@ -376,7 +376,8 @@ func (ot *objectTransformer) objectTransformXML(in interface{}, modifier pathMod
 		path = modifier(path)
 	}
 
-	// For the object use a transform if it exists, if the transform does not find a node it will return nil
+	// For the object use a transform if it exists, if the transform does not find a node it will return nil unless a
+	// default value is specified in which case the default value will be returned
 	if ot.transforms != nil {
 		rawValue, err := ot.transforms.transform(in, "object", modifier, ot.format)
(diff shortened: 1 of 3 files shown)


transform/instance.go
```diff
@@ -390,8 +392,8 @@ func (ot *objectTransformer) objectTransformXML(in interface{}, modifier pathMod
 			if len(v) > 0 {
 				in = v[0]
 			}
-		default:
-			return nil, errors.New("non xml node returned from object transform")
+		case nil:
+			foundTransformObjectNode = false
```
tkuhlman: I don't see how using this bool is effectively any different than just `return ot.defaultValue, nil`

transform/instance.go
```diff
@@ -369,20 +369,24 @@ func (ot *objectTransformer) objectTransformJSON(in interface{}, modifier pathMo
...
 
-	// For the object use a transform if it exists or the default or an empty map
+	// For the object use a transform if it exists, if the transform does not find a node it will return nil
 	if ot.transforms != nil {
 		rawValue, err := ot.transforms.transform(in, "object", modifier, ot.format)
 		if err != nil {
 			return nil, err
 		}
 
+		if rawValue == nil {
```
tkuhlman: Looks good, my only thought is what happens if the transform comes up empty but a default value is set. I think on the JSON side in such a case the default value is used for the object value and then the children transformations proceed.

PR #58, opened 2019-06-21
APIS-2900 Adding current time operation to jstransform

kvincent2:


paas-api-pr-manager:

* [APIS-2900](https://jira.gannett.com/browse/APIS-2900): Create the "current time" operation for jstransform
```
To support _AggregatedTime_ and _LastModified_ fields in XML transformations

*AC*
Add a jstransform operation which returns the result of  
{code:go}
time.Now().Format(time.RFC3339)
{code} as a _string_

```

transform/operations.go
```diff
@@ -246,6 +246,23 @@ func (t *timeParse) transform(raw interface{}) (interface{}, error) {
 	return parsedTime.Format(t.Args["layout"]), nil
 }
 
+type currentTime struct {
+	Args map[string]string
```
abaehre: lower case.

transform/operations.go
```diff
@@ -246,6 +246,23 @@ func (t *timeParse) transform(raw interface{}) (interface{}, error) {
...
+
+func (c *currentTime) init(args map[string]string) error {
+	if err := requiredArgs([]string{"format"}, args); err != nil {
+		return err
+	}
+	c.Args = args
+	return nil
+}
+
+//Errors to worry about?
```
abaehre: you can probably remove this.

transform/operations_test.go
```diff
@@ -62,6 +63,19 @@ func runOpTests(t *testing.T, opType func() transformOperation, tests []opTests)
 		switch {
 		case test.wantErr && err != nil:
 			continue
+		case strings.Contains(test.description,"currentTime") == true:
```
abaehre: no need for == true

transform/operations_test.go
```diff
@@ -62,6 +63,19 @@ func runOpTests(t *testing.T, opType func() transformOperation, tests []opTests)
...
+			now := time.Now()
+			result, ok := test.want.(string)
+			if !ok {
+				t.Error("want must be type string")
+			}
+			wantParse, err := time.Parse(time.RFC3339, result)
+			if err != nil {
+				t.Error(err)
+			}
+			if result := compareTimeStamps(now, wantParse);result != true {
```
abaehre: !result and linting.

transform/operations_test.go
```diff
@@ -374,3 +388,34 @@ func TestTimeParse(t *testing.T) {
...
+			wantInitErr: true,
+		},
+	}
+	runOpTests(t, func() transformOperation { return &currentTime{} }, tests)
+}
+
+func compareTimeStamps (time1 time.Time, time2 time.Time) (bool) {
+	maxTimeDifference := time.Duration(1200) * time.Second
+	actualDiff := time1.Sub(time2)
+	if actualDiff > maxTimeDifference {
```
abaehre: return maxTimeDifference > actualDiff

also 1200 seems arbitrary?

transformSchema.json
```diff
@@ -272,6 +275,34 @@
 						}
 					}
 				}
+			},
+			"currentTime" : {
+				"description": "Returns the current time formatted as a string",
+				"type": "object",
+				"required": [
```
abaehre: you missed type

transform/operations_test.go
abaehre: also this seems sketchy. I think it'd be better if you didn't use runOpTests.

transform/operations.go
kvincent2: why is this one lower case when all of the other operations are upper case?

transform/operations_test.go
```diff
@@ -58,10 +60,24 @@ func runOpTests(t *testing.T, opType func() transformOperation, tests []opTests)
 		}
 
 		got, err := op.transform(test.in)
+		fmt.Println(got)
```
abaehre: this needs to get removed.

abaehre:
@kvincent2 for some reason can't comment on your comment, but these should all be lowercase. It was an oversight from previous users.

transform/operations.go
abaehre: commented below. but these should all be lower case. probably an oversight.

transform/operations_test.go
abaehre: missed the == true. 

And to the point - you're cluttering up a common test function a lot with test specific logic. It should get moved out, or somehow otherwise leveraged / modified.

transform/operations_test.go
```diff
@@ -375,21 +400,74 @@ func TestTimeParse(t *testing.T) {
...
 		},
-
 	}
 	runOpTests(t, func() transformOperation { return &stringToInteger{} }, tests)
 }
+
+func compareTimeStamps(time1 time.Time, time2 time.Time) bool {
+	maxTimeDifference := time.Duration(300) * time.Second
+	actualDiff := time1.Sub(time2)
+	return actualDiff < maxTimeDifference
```
ahornerr: Might want to take the absolute value of `actualDiff` in case it's negative

transform/operations_test.go
```diff
@@ -375,21 +400,74 @@ func TestTimeParse(t *testing.T) {
...
+			wantInitErr: true,
+		},
+		{
+			description: "Too many args",
+			args:        map[string]string{"format": time.RFC3339, "cookies": "failure"},
+			wantInitErr: true,
+		},
+	}
+
+	for _, test := range tests {
```
ahornerr: Use `t.Run(test.description, ...` inside the loop and replace your `t.Error()`s in the loop with `t.Fatal()`

transform/operations_test.go
```diff
@@ -45,31 +46,55 @@ type opTests struct {
...
 
-		got, err := op.transform(test.in)
+func runOpTestInit(opType func() transformOperation, test opTests) (transformOperation, error) {
+	op := opType()
+	err := op.init(test.args)
+	switch {
+	case test.wantInitErr && err != nil:
+		return op, err
+	case test.wantInitErr && err == nil:
+		return op, fmt.Errorf("Test %q - got init error nil, want error", test.description)
```
ahornerr: Remove `test.description` from the error message, my comment about `t.Run()` below will add it automatically

transform/operations_test.go
```diff
@@ -45,31 +46,55 @@ type opTests struct {
...
+	case !test.wantErr && err != nil:
```
ahornerr: I feel like there's a lot of duplicated code in the error vs wantErr checking. It can be pulled out pretty easily into a new method that accepts two `error`s as arguments and returns an `error`

transform/operations_test.go
```diff
@@ -45,31 +46,55 @@ type opTests struct {
...
-		}
+func runOpTestTransform(op transformOperation, test opTests) (interface{}, error) {
+	got, err := op.transform(test.in)
+
+	switch {
+	case test.wantErr && err != nil:
+		return got, err
+	case test.wantErr && err == nil:
+		return got, fmt.Errorf("Test %q - got nil, want error", test.description)
+	case !test.wantErr && err != nil:
```
ahornerr: Neither this or the above case actually compares the errors, just that there WAS an error. It would be better if we compared the `err.Error()` values against each other to make sure we're getting the *right* error

transform/operations_test.go
```diff
@@ -45,31 +46,55 @@ type opTests struct {
...
+func runOpTest(t *testing.T, opType func() transformOperation, test opTests) {
+	op, err := runOpTestInit(opType, test)
+	if test.wantInitErr && err != nil {
+		return
+	} else if err != nil {
+		t.Error(err)
+	}
+
+	_, err = runOpTestTransform(op, test)
+	if test.wantErr && err != nil {
```
ahornerr: See comments below

transform/operations.go
```diff
@@ -239,11 +239,32 @@ func (t *timeParse) transform(raw interface{}) (interface{}, error) {
...
+	}
+	c.args = args
+	return nil
+}
+
+func (c *currentTime) transform(raw interface{}) (interface{}, error) {
+	timeFmt := c.args["format"]
+	switch c.args["format"] {
+	case "RFC3339":
+		timeFmt = time.RFC3339
```
ahornerr: 💯

transform/operations_test.go
```diff
@@ -375,21 +400,108 @@ func TestTimeParse(t *testing.T) {
...
+			description: "Format not predefined",
+			args:		 map[string]string{"format": "Mon Jan 2 15:04:05 MST 2006"},
+			want:		time.Now().Format("Mon Jan 2 15:04:05 MST 2006")	,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.description, func(t *testing.T) {
+			op, err := runOpTestInit(func() transformOperation { return &currentTime{} }, test)
+			if err != nil && test.wantInitErr {
```
ahornerr: Don't you have `runOpTestInit` which can do all this error checking and stuff already?

transform/operations_test.go
```diff
@@ -375,21 +400,108 @@ func TestTimeParse(t *testing.T) {
...
+			}
+			now := time.Now()
+			result, ok := got.(string)
+			if !ok {
+				t.Fatal("function must return string")
+			}
+
+			var gotParse time.Time
+
+			if test.args["format"] == "RFC3339" {
```
ahornerr: It would be better to put the expected time in the test case rather than calculating it down here

transform/operations.go
abaehre: I'm confused... by this. Format takes a string. so why all this extra stuff when what you actually want is to type assert the format arg to a string and then use it?

jstorer: bc passing in "RFC3339" as an argument cant be used directly as its meaningless. the time.Format() just takes it as a string instead of grabbing the const from the time package. So we want means to pass in the typical go time formatting date as well as shorthand versions on a case by case basis

abaehre: right... but I woudl pass in time.RFC3339 as the argument. not RFC3339

ahornerr: It's just a convenience method for clarity rather than having to actually find the constant value for `time.RFC3339` and reading over the transforms and understanding that's what it is. 

https://golang.org/pkg/time/
```golang
        RFC3339     = "2006-01-02T15:04:05Z07:00"
```

abaehre: like. RFC3339 is the name of a format not an actual format.

transform/operations_test.go
kvincent2: Working on this. I see what you mean.

transform/operations_test.go
```diff
@@ -45,33 +46,55 @@ type opTests struct {
...
-		case !reflect.DeepEqual(got, test.want):
-			t.Errorf("Test %q - got\n%s\nwant\n%s", test.description, got, test.want)
-		}
+		runOpTest(t, opType, test)
 	}
 }
 
+func runOpTest(t *testing.T, opType func() transformOperation, test opTests) {
+	op, err := runOpTestInit(opType, test)
+	if test.wantInitErr && err != nil {
```
ahornerr: This doesn't test the case in which we want an error but none is thrown

transform/operations_test.go
```diff
@@ -45,33 +46,55 @@ type opTests struct {
...
+		return
+	} else if err != nil {
+		t.Error(err)
+	}
+
+	_, err = runOpTestTransform(op, test)
+	if test.wantErr && err != nil {
```
ahornerr: Same as above

transform/operations_test.go
ahornerr: This still needs resolved.

Something like

```golang
wantTime, err := time.Parse(test.args["format"], wantResult)
gotTime, err := time.Parse(test.args["format"], gotResult)
```

then compare the two

ahornerr: Or change your `test.want` to be a time object instead of a string and then you only have to do the time.Parse once

transform/operations.go
abaehre: since we are doing it via a config in json we can't do it like that. this is correct just sad.

transform/operations.go
```diff
@@ -239,11 +239,32 @@ func (t *timeParse) transform(raw interface{}) (interface{}, error) {
...
+	if err := requiredArgs([]string{"format"}, args); err != nil {
+		return err
+	}
+	c.args = args
+	return nil
+}
+
+func (c *currentTime) transform(raw interface{}) (interface{}, error) {
+	timeFmt := c.args["format"]
+	switch c.args["format"] {
```
mmandolesi-g: Would you want a `default` case to this switch statement?

transform/operations_test.go
```diff
@@ -375,21 +386,122 @@ func TestTimeParse(t *testing.T) {
...
+				return
+			}
+			got, err := runOpTestTransform(op, test)
+
+			if err != nil {
+				t.Fatal(err)
+				return
+			}
+
+			if test.want == nil {
```
abaehre: why is this case needed?

transform/operations_test.go
```diff
@@ -375,21 +386,122 @@ func TestTimeParse(t *testing.T) {
...
+
+			gotResult, ok := got.(string)
+			if !ok {
+				t.Fatal("function must return string")
+			}
+
+			var gotParse time.Time
+			var wantParse time.Time
+
+			if test.args["format"] == "RFC3339" {
```
abaehre: I think you handled this in your function so you shouldn't need this if else.

transform/operations_test.go
```diff
@@ -375,21 +386,122 @@ func TestTimeParse(t *testing.T) {
...
+					t.Fatal(err)
+				}
+			} else {
+				gotParse, err = time.Parse(test.args["format"], gotResult)
+				wantParse, err = time.Parse(test.args["format"], wantResult)
+				if err != nil {
+					t.Fatal(err)
+				}
+			}
+			if result := compareTimeStamps(wantParse, gotParse); !result {
```
abaehre: !compareTimeStamps(wantParse, gotParse)

transform/operations_test.go
```diff
@@ -375,21 +386,122 @@ func TestTimeParse(t *testing.T) {
...
+			wantErr:     true,
 		},
-
 	}
 	runOpTests(t, func() transformOperation { return &stringToInteger{} }, tests)
 }
+
+func compareTimeStamps(time1 time.Time, time2 time.Time) bool {
+	maxTimeDifference := time.Duration(300) * time.Second
+	actualDiff := absValue(time1.Sub(time2))
```
abaehre: if you flip time2 and time 1 you shouldn't be absValue (which I'm pretty sure exists in math though it takes a float64)

transform/operations_test.go
```diff
@@ -375,21 +386,122 @@ func TestTimeParse(t *testing.T) {
...
+func compareWantErrs(gotErr error, wantErr bool) error {
+	switch {
+	case wantErr && gotErr == nil:
+		return errors.New("got nil, want error")
+	case wantErr && gotErr != nil:
+		return nil
+	case !wantErr && gotErr == nil:
+		return nil
+	case !wantErr && gotErr != nil:
+		return errors.New(fmt.Sprintf("got error, want nil: %v", gotErr))
```
abaehre: this could just be fmt.Errorf()

transform/operations_test.go
```diff
@@ -375,21 +386,122 @@ func TestTimeParse(t *testing.T) {
...
+		t.Run(test.description, func(t *testing.T) {
+			op, err := runOpTestInit(func() transformOperation { return &currentTime{} }, test)
+			if err != nil {
+				return
+			}
+			got, err := runOpTestTransform(op, test)
+
+			if err != nil {
+				t.Fatal(err)
+				return
```
abaehre: fatal will close this out.

transform/operations_test.go
```diff
@@ -375,21 +386,122 @@ func TestTimeParse(t *testing.T) {
...
+			args:        map[string]string{"format": "Mon Jan 2 15:04:05 MST 2006"},
+			want:        time.Now().Format("Mon Jan 2 15:04:05 MST 2006"),
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.description, func(t *testing.T) {
+			op, err := runOpTestInit(func() transformOperation { return &currentTime{} }, test)
+			if err != nil {
+				return
```
abaehre: shouldn't we fatal out if we get an error?

abaehre: and then shouldn't need a return.

transform/operations_test.go
```diff
@@ -43,35 +43,46 @@ type opTests struct {
...
+	if err := compareWantErrs(err, test.wantInitErr); err != nil {
+		return op, err
+	}
+	return op, nil
+}
+
+func runOpTestTransform(op transformOperation, test opTests) (interface{}, error) {
+	got, err := op.transform(test.in)
+
+	if err := compareWantErrs(err, test.wantErr); err != nil {
```
abaehre: this is simpler as 
`return got, compareWantErrs(err, test.wantErr)`

transform/test_data/operations.json
```diff
@@ -137,6 +137,26 @@
         }
       }
     },
+    "lastModified": {
```
abaehre: shouldn't there be equivalent stuff added to the transform_test file?

transformSchema.json
```diff
@@ -1,502 +1,556 @@
...
+},
+"type": {
+"anyOf": [
+{
+"$ref": "#/definitions/simpleTypes"
+},
+{
+"type": "array",
+"items": {
+"$ref": "#/definitions/simpleTypes"
```
abaehre: you kinda... shredded the formatting on this.

transform/operations_test.go
kvincent2: Because if this isn't there, my test (and only my test) tries to do the transform, even when the init is designed to fail.

kvincent2: For more detail:
Other tests call 
```
// A common test runner for all the operations tests
func runOpTests(t *testing.T, opType func() transformOperation, tests []opTests) {
	for _, test := range tests {
		t.Run(test.description, func(t *testing.T) {
			runOpTest(t, opType, test)
		})
	}
}

func runOpTest(t *testing.T, opType func() transformOperation, test opTests) {
	op, err := runOpTestInit(opType, test)
	if err != nil {
		t.Fatal(err)
	}
	if test.wantInitErr {
		return
	}
	_, err = runOpTestTransform(op, test)
	if err != nil {
		t.Fatal(err)
	}
}
```

Which stops processing if init fails. But my test calls them individually.

abaehre: Right, but if you fatal out when the error returns on runOpTestInit then it won't get to that point right?

transform/operations.go
abaehre: hypothetically the default is rfc3339. Anything else is up to the peoples.

transform/operations_test.go
```diff
@@ -405,6 +406,33 @@ func TestToCamelCase(t *testing.T) {
...
+		{
+			description: "Too many args",
+			args:        map[string]string{"format": "RFC3339", "cookies": "failure"},
+			wantInitErr: true,
+		},
+		{
+			description: "Format not predefined",
+			args:        map[string]string{"format": "Mon Jan 2 15:04:05 MST 2006"},
+			want:        time.Now().Format("Mon Jan 2 15:04:05 MST 2006"),
+		},
```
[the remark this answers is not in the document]
abaehre: You did in the other test file, but just to be clear in our unit tests.

transform/operations_test.go
```diff
@@ -405,6 +406,33 @@ func TestToCamelCase(t *testing.T) {
...
+	}
+
+	runOpTests(t, func() transformOperation { return &currentTime{} }, tests)
```
abaehre: what was the point of splitting everything out if you never use that special use case for checking if the time is within a certain time limit?

transform/operations_test.go
```diff
@@ -43,32 +44,31 @@ type opTests struct {
...
-			t.Errorf("Test %q - got error, want nil: %v", test.description, err)
-		case !reflect.DeepEqual(got, test.want):
-			t.Errorf("Test %q - got\n%s\nwant\n%s", test.description, got, test.want)
-		}
+	if err := compareWantErrs(err, test.wantErr); err != nil {
+		t.Fatal(err)
+	}
+
+	if !reflect.DeepEqual(got, test.want) {
+		t.Fatalf("got: %s, want: %s", got, test.want)
```
ahornerr: I liked the newlines :(

PR #59, opened 2019-06-25
APIS-2816 - Adding a String to Int operation to jstransform

adilbozai1995:


paas-api-pr-manager:

* [APIS-2816](https://jira.gannett.com/browse/APIS-2816): Add a String to Int operation to jstransform 
```
In some cases we may want to cast the result of the of the transformation to integer.
 For example, we may need to perform a _replace_ operation on *id*, and then cast the result of _replace_ to integer (_replace_ returns a string)
{noformat}
"id": {
  "type": "string",
  "transform": {
    "sport": {
      "from": [
        {
          "xmlPath": "id",
          "operations": [
            {
              "type": "replace",
              "args": {
                "regex": "[\\D]",
                "new": ""
              }
            }
          ]
        }
      ]
    }
  }
}
{noformat}
For that we may delevlop a new jstransform operation, so the trasformation will look like
{noformat}
"operations": [
    {
      "type": "replace",
      "args": {
        "regex": "[\\D]",
        "new": ""
      }
    },
    {
      "type": "atoi"
    }
  ]
{noformat}
*AC*
 develop a new _jstransform_ operation that accepts a string, tries to cast it to an integer and returns this integer
```

transform/operations.go
```diff
@@ -246,6 +246,28 @@ func (t *timeParse) transform(raw interface{}) (interface{}, error) {
...
+
+func (s *stringToInteger) transform(raw interface{}) (interface{}, error) {
+	str, ok := raw.(string)
+	if !ok {
+		return nil, errors.New("stringToInteger only supports strings")
+	}
+
+	result, err := strconv.Atoi(str)
+	if err != nil {
+		return nil, errors.New(fmt.Sprintf("error converting to an Integer from a string: %s", err))
```
abaehre: if you're going to use fmt and errors you can probably just use fmt.Errorf()

abaehre: actually you probably don't need most of this

you could just 
`return strconv.Atoi(str)`

transform/operations_test.go
```diff
@@ -374,3 +374,28 @@ func TestTimeParse(t *testing.T) {
...
+	tests := []opTests{
+
+		{
+			description: "Simple working case",
+			in: "237754",
+			want: 237754,
+		},
+
+		{
+			description: "An alphabet input",
```
abaehre: alpha*

transform/operations_test.go
```diff
@@ -374,3 +374,28 @@ func TestTimeParse(t *testing.T) {
...
+			in: "a",
+			wantErr: true,
+		},
+
+		{
```
abaehre: don't really need this case.

PR #61, opened 2019-06-25
APIS-2819 - Added an operation function to convert a string with dashes to a camel case string

fcantor:


paas-api-pr-manager:

* [APIS-2819](https://jira.gannett.com/browse/APIS-2819): Add a new dashes-to-CamelCase operation to jstransform
```
There is a need to transform strings with dashes to camelCase.

For example in the

[https://github.com/GannettDigital/sports-aggregator/blob/bd79df802f048fce8321bf6698f4301e6f60d35d/processor/sdiprocessor/test_data/boxscores/boxscores_MLB.xml#L1660] we need to extract "extra-base-hit" from id and then transform it to "extraBaseHit".

We may want to develop a new jstransform operation for that.

*AC*

Develop a new jstransform operation that accepts a string, capitalize letters after dashes, removes dashes  and returns the resulting string

example:
 * xxx-yyy-zzz - > zzzYyyZzz
 * xxx -> xxx
```

PR #62, opened 2019-06-25
APIS-2816 - added the stringToInteger operation in the Operations Table in transform.adoc

adilbozai1995:


paas-api-pr-manager:

* [APIS-2816](https://jira.gannett.com/browse/APIS-2816): Add a String to Int operation to jstransform 
```
In some cases we may want to cast the result of the of the transformation to integer.
 For example, we may need to perform a _replace_ operation on *id*, and then cast the result of _replace_ to integer (_replace_ returns a string)
{noformat}
"id": {
  "type": "string",
  "transform": {
    "sport": {
      "from": [
        {
          "xmlPath": "id",
          "operations": [
            {
              "type": "replace",
              "args": {
                "regex": "[\\D]",
                "new": ""
              }
            }
          ]
        }
      ]
    }
  }
}
{noformat}
For that we may delevlop a new jstransform operation, so the trasformation will look like
{noformat}
"operations": [
    {
      "type": "replace",
      "args": {
        "regex": "[\\D]",
        "new": ""
      }
    },
    {
      "type": "atoi"
    }
  ]
{noformat}
*AC*
 develop a new _jstransform_ operation that accepts a string, tries to cast it to an integer and returns this integer
```

PR #68, opened 2019-07-02
APIS-2964 string to float operation

johnkuhlin:


paas-api-pr-manager:

* [APIS-2964](https://jira.gannett.com/browse/APIS-2964): Add a String to Float operation to jstransform
```
In some cases we may want to cast the result of the of the transformation to float.
For example, we may need to perform a replace operation on games behind, and then cast the result of replace to float (replace returns a string)

AC
develop a new jstransform operation that accepts a string, tries to cast it to a float and returns this value
```

transform/operations.go
```diff
@@ -318,6 +318,30 @@ func (s *stringToInteger) transform(raw interface{}) (interface{}, error) {
...
+// stringToFloat is a transformOperation which takes a string and converts it into a float
+type stringToFloat struct {
+}
+
+func (s *stringToFloat) init(args map[string]string) error {
+	return nil
+}
+
+func (s *stringToFloat) transform(raw interface{}) (interface{}, error) {
+	_, ok := raw.(float64)
```
abaehre: is this first bit necessary?

transform/operations_test.go
```diff
@@ -456,6 +456,37 @@ func TestStringToInteger(t *testing.T) {
...
+			in:          "11.0",
+			want:        float64(11),
+		},
+		{
+			description: "no decimal",
+			in:          "13",
+			want:        float64(13),
+		},
+		{
+			description: "Boolean inpu",
```
abaehre: input

transform/operations.go
johnkuhlin: It is. The tests keep failing. Took a while before I realized that raw was already pass in as float.

PR #69, opened 2019-07-02
add support for integer type in convert() and remove strToInt and strToFloat as they are redundant and can be done without operations

jstorer:


paas-api-pr-manager:

* [APIS-2916](https://jira.gannett.com/browse/APIS-2916): content-api-sport schema organization for xml transforms
```
as xml transforms have progressed for SDI we are ending up with a lot of duplicate schema that is living in sports-aggregator. This schema in aggregator should be moved into capi-sport with the transforms placed upon the output schema we already have whenever possible to mitigate duplication.

The output schema itself also needs analyzed and reorganized to best fit xml transforms without altering its overall structure (as in when you run go generate it still outputs what is already there)

AC
- move transformSchema from sports-agg to capi-sport

- reuse output schema when possible

- analyze structure of output schema and organize in such a way to best leverage xml transforms
```

PR #71, opened 2019-07-26
APIS-3051 - Add support for generating Avro schema files

tkuhlman:


paas-api-pr-manager:

* [APIS-3051](https://jira.gannett.com/browse/APIS-3051): Add support to jstransform for generating Avro schemas from JSON Schemas
```
Findings from the PoC APIS-2982 have led us to discover that we need a tool to generate Avro schemas from JSON Schemas. jstransform already has a bunch of the helper methods and generates Go structs from JSON Schemas so it's an obvious place for this functionality to live.

For generating coachespoll Avro schemas I used a Node.js package that has most of the functionality we need but we would like to rewrite it in Go. [https://www.npmjs.com/package/jsonschema-avro]

Example usage of the jsonschema-avro package [https://github.com/GannettDigital/content-api-sport/blob/d297c8812ff27dff950d73d66c15c4640020ff0d/avro.js]

AC:
 * Add generation of Avro schemas from JSON schema to jstransform
 * Add a test to ensure that the schemas have been generated (like JSON schema to go struct generation tests)
 * Ensure that this generation can resolve $refs
 * Ensure that this generation can support JSON schema allOfs
 * Our data doesn't work with nullable Avro values (union of size 2 where the first type is null) so we should just assume no fields are nullable and have a default value for that type (int = 0, string = "", etc.)

 
```

tkuhlman:
A quick note everything in struct* is not new just moved. Previously it all lived in generate.go and I thought it was cleaner to have generate.go for the high level and common stuff, struct.go for the struct generation code and avro.go for the avro generation code

README.md
```diff
@@ -31,6 +31,14 @@ or if you have compiled the tool and have it in your path rather than vendoring
 
 then simply run `go generate`.
 
+As well as generating Go structs serialization libraries for [Message Pack](https://msgpack.org/) and [Avro](https://avro.apache.org/) are available.
+For Avro the schema as well as serialization and helper functions are generated.
+The Avro schema also adds two metadata fields, `AvroWriteTime` and `AvroDeleted`.
```
ahornerr: 👍

generate/avro.go
```diff
@@ -0,0 +1,304 @@
...
+type avroConfig struct {
+	dir       string
+	namespace string
+	writer    io.Writer
+}
+
+// buildAvroSerializationFunctions will run gogen-avro on the given schema to build a serialization and deserialization
+// code for the schema.
+func buildAvroSerializationFunctions(outputDir string) error {
+	// TODO buildAvroSerializationFunctions - this should take the schema and run gogen-avro
```
ahornerr: Is this something that will be done in another ticket/PR?

generate/avro.go
```diff
@@ -0,0 +1,304 @@
...
+	return nil
+}
+
+// buildAvroSchemaFile will generate an Avro schema based on the go struct with the given name in the file at the
+// given path. I will write it to the same directory as the go file as `<name>.avsc` but with the name lowercased.
```
ahornerr: 👍

generate/avro.go
```diff
@@ -0,0 +1,304 @@
...
+// http://avro.apache.org/docs/current/spec.html
+//
+// By default the file is written with now whitespace to minimize the size, choose pretty for better formatting
+//
+// Note: Avro can't handle maps with a key other than a string, http://avro.apache.org/docs/current/spec.html#Maps
+// Neither can JSON schema, https://json-schema.org/understanding-json-schema/reference/object.html so this only
+// becomes relevant if it is used with go structs which weren't just generated from JSON schema
+func buildAvroSchemaFile(name, dir string, spec *ast.TypeSpec, pretty bool) (string, error) {
```
ahornerr: I find it interesting that we're building the Avro schema based off of the Go structs and not the JSON Schema but I did see your comment about why that decision was made and I'm happy with this solution too

generate/avro.go
```diff
@@ -0,0 +1,304 @@
...
+		dir:    dir,
+		writer: writer,
+	}
+
+	astutil.Apply(spec, writeAvroStructPre(cfg, name, avroHeader), writeAvroFieldPost(writer))
+
+	if pretty {
+		rawJSON := json.RawMessage(buf.Bytes())
+		enc := json.NewEncoder(specFile)
+		enc.SetIndent("", "  ")
```
ahornerr: 👍

generate/avro.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	if err := specFile.Close(); err != nil {
+		return "", fmt.Errorf("failed closing file: %v", err)
+	}
+
+	return outPath, nil
+}
+
+// parseGoStruct parses the go file at path returning the named struct type definition as an *ast.TypeSpec
```
ahornerr: This is interesting, didn't know you could do this

generate/avro.go
```diff
@@ -0,0 +1,304 @@
...
+		case "string":
+			typeName = "string"
+		default:
+			typeName = "unknown"
+		}
+		if nullable {
+			if typeName == "boolean" {
+				return fmt.Sprintf(`"type":%q,"default":false`, typeName)
+			}
+			return fmt.Sprintf(`"type":["null",%q]`, typeName)
```
ahornerr: So one thing to consider, our data doesn't line up with this and will cause a serialization error when we try to convert the `map[string]interface{}` to Avro data. Perhaps this isn't an issue with the avro-gen library but the Linkedin one I was using previously was complaining. In order to work around this, I set every nullable field as not nullable with a default value for the type.

generate/avro.go
```diff
@@ -0,0 +1,304 @@
...
+	"path/filepath"
+	"reflect"
+	"strings"
+
+	"golang.org/x/tools/go/ast/astutil"
+)
+
+const (
+	avroHeader           = `{"name":%q,"type":"record","fields":[{"name":"AvroWriteTime","doc":"The timestamp when this avro data is written. Useful for identifying the newest row of data sharing keys.","type":"long","logicalType":"timestamp-millis"},{"name":"AvroDeleted","doc":"This is set to true when the Avro data is recording a delete in the source data.","default":false,"type":"boolean"},`
+	embeddedStructHeader = `{"name":%q,"type":"record","fields":[`
```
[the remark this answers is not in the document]
tkuhlman: I looked into using some of the code from a couple of Avro libraries but nothing I found was really built for this purpose. Additionally it seemed best not to tie us to a particular Avro library in step 1 of this process.

I then considered a few different ways of generating this schema but ultimately decided on parsing the Go struct for two reasons. First, I thought it would result in simpler more clear code to just parse that and print out the JSON for the Avro schema than it would to first build up an internal representation then write out the schema based on that. Now that it is done I'm not sure it is any clearer. The second was that I knew I was going to need to parse the go structs in step 3 of this process in order to generate helper functions for processing the code. I thought having some code that helps with parsing go structs already built would help then.

generate/avro.go
tkuhlman: Yes another PR or two is on its way. I think there is a 3 step process for the Avro code generation, step 1 is the Avro schema, step 2 is to make the Avro serialization/deserialization code, step 3 is to add helpers that make writing out Avro files easy for each data type.

generate/avro.go
tkuhlman: Interesting I was using the Google documentation and the Avro schema documentation as guide and both recommend this as the approach for nullable fields.
https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-avro
http://avro.apache.org/docs/current/spec.html

ahornerr: You're correct in that it's the recommended approach the problem is that our data doesn't fit this format. Rather than transforming our data to fit the format it's easier to specify the fields as non-nullable in the schema and any missing fields will have their defaults applied

generate/avro.go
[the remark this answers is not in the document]
ahornerr: Sounds good thanks for the explanation

generate/avro.go
```diff
@@ -0,0 +1,344 @@
...
+type avroConfig struct {
+	dir           string
+	excludeFields map[string]bool
+	namespace     []string
+	writer        io.Writer
+}
+
+// buildAvroSerializationFunctions will run gogen-avro on the given schema to build a serialization and deserialization
+// code for the schema. It will create the directory `avro/<schema name>` and add the generated files there. The directory
+// is necessary because generated code for different types will clobber each other.
```
ahornerr: Makes sense

generate/avro.go
```diff
@@ -0,0 +1,344 @@
...
+	tag := reflect.StructTag(strings.Trim(literal.Value, "`"))
+
+	description = tag.Get("description")
+	jsonValue := tag.Get("json")
+	jsonSplits := strings.Split(jsonValue, ",")
+	name = jsonSplits[0]
+	if len(jsonSplits) > 1 {
+		for _, split := range jsonSplits[1:] {
+			if strings.ToLower(split) == "omitempty" {
+				omitEmpty = true
```
ahornerr: Clever use of named return values

generate/avro.go
```diff
@@ -0,0 +1,344 @@
...
+// writeAvroStruct returns an apply function intended to be called for the start of each node.
+// The corresponding Post function will be called after the children of the node have been traversed.
+func writeAvroStruct(cfg avroConfig, name, defaultFields string) astutil.ApplyFunc {
+	return func(c *astutil.Cursor) bool {
+		if c.Name() == "Node" {
+			var namespace string
+			if len(cfg.namespace) != 0 {
+				namespace = fmt.Sprintf(`"namespace":%q,`, strings.Join(cfg.namespace, "."))
+			}
+			if _, err := fmt.Fprintf(cfg.writer, `"name":%q,%s"type":"record","fields":[%s`, name, namespace, defaultFields); err != nil {
```
ahornerr: Is there some way we can report this error?

generate/avro.go
```diff
@@ -0,0 +1,344 @@
...
+}
+
+// writeAvroStruct returns an apply function intended to be called for the start of each node.
+// The corresponding Post function will be called after the children of the node have been traversed.
+func writeAvroStruct(cfg avroConfig, name, defaultFields string) astutil.ApplyFunc {
+	return func(c *astutil.Cursor) bool {
+		if c.Name() == "Node" {
+			var namespace string
+			if len(cfg.namespace) != 0 {
+				namespace = fmt.Sprintf(`"namespace":%q,`, strings.Join(cfg.namespace, "."))
```
ahornerr: Perhaps rename the `namespace` variable for clarity so that it's obvious it includes json formatting

generate/avro_test.go
```diff
@@ -0,0 +1,202 @@
...
+			description: "with embedded and nested structs, fields with descriptions",
+			name:        "Complex",
+			goPath:      "./test_data/complex.go.out",
+			wantPath:    "./test_data/complex.avsc.out",
+		},
+	}
+
+	// For the complex.go to find the embedded struct simple.go must exist, other tests use that filename so create
+	// and destroy it here
+	if err := os.Link("./test_data/simple.go.out", "./test_data/simple.go"); err != nil {
```
ahornerr: I wonder if this works on Windows

generate/generate.go
```diff
@@ -131,6 +141,40 @@ func BuildStructsWithArgs(args BuildArgs) error {
...
+	}
+
+	// step 2 build serialization functions from the new avro schema
+	// this step parses the just created Avro schema and by doing so acts as validation step as well
+	if err := buildAvroSerializationFunctions(avroSchemaPath); err != nil {
+		return fmt.Errorf("failed to build Avro serialization code: %v", err)
+	}
+
+	// step 3 generate helper functions
+	// TODO generateAvroHelpers - includes code to convert from the main struct to avro struct and vice versa, bulk and individual avro writers that will write out the avro to an io.writer setting the Avro metadata fields also, a AvroDeleted method
```
ahornerr: Looking forward to seeing what you come up with for this

ahornerr: Also good use of comments 👍

generate/avro.go
tkuhlman: Not without making it some sort of type and making this a method of that type which seems pretty complicated. What will end up happening right now is the JSON will be invalid and fail later parsing.

PR #72, opened 2019-08-01
APIS-3185 Added ability to pick up root level properties when allOf/oneOf is used

ahornerr:


paas-api-pr-manager:

* [APIS-3185](https://jira.gannett.com/browse/APIS-3185): Fix jstransform generate bug allOf/oneOf required to generate
```
There exists an issue in jstransform where structs won't be generated unless their source JSON Schema is contained inside an allOf/oneOf block. This behavior is undesired and we would like to generate structs even if they don't contain allOf/oneOf directives.

[https://github.com/GannettDigital/jstransform/blob/master/generate/generate.go#L73]

Current behavior:

Omitting this line to generate structs from entry.json [https://github.com/GannettDigital/content-api-sport/blob/master/model/coachesPoll/generator.go#L3] will cause all of the proceeding structs to not be generated (ballot, poll, coach, etc). This is because entry.json contains a oneOf for each of the types (ballot, poll, coach, etc) but the other types only contain an allOf for base.json. In this case, base.go struct will be generated but nothing else.

Desired behavior is for all of the types plus base to be generated even without entry.json.

AC:
 * Fix the above described behavior
 * Add tests
```

jsonschema/schema.go
```diff
@@ -105,15 +105,15 @@ func SchemaFromFile(schemaPath string, oneOfType string) (*Schema, error) {
 }
 
 // SchemaTypes will parse the given file and report which top level allOfTypes and oneOfTypes are found in that schema.
```
tkuhlman: Update the comment to let users of this know what the 3rd return value is.

ahornerr: Thanks!

PR #73, opened 2019-08-01
Add Avro serialization and helper functions for using it

tkuhlman:


tkuhlman:
Testing needs to be extended, I will be following up with that in the next PR.

PR #76, opened 2019-08-07
Add a generated function for bulk writing to an Avro Container File

tkuhlman:


generate/avro_template.go
```diff
@@ -93,6 +130,33 @@ func Test{{ .name }}_WriteAvroCF(t *testing.T) {
...
+
+	devnull, _ := os.Open("/dev/null")
+	defer devnull.Close()
+
+	errChan := {{ .name }}BulkAvroWriter(devnull, time.Now(), inputChan)
+
+	for _, item := range input {
+		select {
+		case err := <-errChan:
+			fmt.Print(err)
```
abaehre: wouldn't it be better to log?

generate/avro_template.go
```diff
@@ -93,6 +130,33 @@ func Test{{ .name }}_WriteAvroCF(t *testing.T) {
 		t.Error("OCF reports deleted true expected false")
 	}
 }
+
+func Example{{ .name }}BulkAvroWrite() {
+	input := []*{{ .name }}{ {}, {}, {} }
+	inputChan := make(chan *{{ .name }})
+
+	devnull, _ := os.Open("/dev/null")
```
abaehre: err?

generate/avro_template.go
```diff
@@ -93,6 +130,33 @@ func Test{{ .name }}_WriteAvroCF(t *testing.T) {
...
+			fmt.Print(err)
+			return
+		case inputChan <- item:
+		}
+	}
+
+	// Check for any final errors, the errorChan should be closed when the BulkWriter is finished processing
+	for err := range errChan {
+		if err != nil {
+			fmt.Print(err)
```
abaehre: log?

generate/avro_template.go
tkuhlman: definitely but given this was an example and for example functions there is no return and there is value in keeping them shorter to just show the framework I skipped it.

generate/avro_template.go
tkuhlman: Yeah but it is an example, so I was just trying to express the minimum do something

PR #78, opened 2019-08-08
Add a function to better handle writing deleted data as Avro files

tkuhlman:
The new function allows populating the object with some data, such as
an ID before writing the Avro file.


generate/avro_template.go
```diff
@@ -5,6 +5,7 @@ var (
 package {{ .pkgName }}
 
 import (
+    "errors"
```
abaehre: lint?

PR #84, opened 2019-09-05
APIS-3408 Allow Avro to support generation of nullable nested objects

ahornerr:


generate/avro.go
```diff
@@ -187,11 +187,27 @@ func filterPackage(name, path string) (*ast.File, error) {
 	}
 	var goFile *ast.File
 	for _, f := range pkg.Files {
-		if len(f.Decls) == 1 {
-			if goFile != nil {
-				return nil, fmt.Errorf("name %q is not unique in go files at path %q", name, path)
+		// It's necessary to loop over all the decls and their specs to ensure the typeSpec.Name.Name matches our name
+		// because the `ast.FilterPackage` doesn't filter out declarations that only have the wanted itemName as a
```
tkuhlman: Interesting it doesn't filter out when the name is an argument. I bet that was a pain to find.

ahornerr: It wasn't too terrible but not immediately obvious what was going on. The `ast` package doesn't give you the ability to easily filter these out like we need unfortunately.

PR #88, opened 2020-02-14
APIS-4283 fix typing errors and messages

brianlieberman:


jsonschema/validator.go
```diff
@@ -61,8 +61,17 @@ func (v *validator) Validate(raw json.RawMessage) (bool, error) {
...
 		sort.Slice(result.Errors(), func(i, j int) bool {
+
 			return result.Errors()[i].String() < result.Errors()[j].String()
 		})
+
+		// make invalid typing user friendly
+		for i := range result.Errors() {
+			if result.Errors()[i].Type() == "invalid_type" && result.Errors()[i].Details()["given"] == "array" {
+				result.Errors()[i].SetDescription("Expected: " +
+					fmt.Sprintf("%v", result.Errors()[i].Details()["expected"]) + " Given: string")
```
tkuhlman: This this error message only happen when the type given was a string? That seems like just the specific instance we recreated and this invalid message can happen for other types also. I wonder if it is best to leave this as is, this fix may cause other issues. The error message lacked a bit of clarity but wasn't fundamentally wrong.

Perhaps it is best to just add context, wrap this error as it comes back from the transformer in a note say that it was an error from validation after transformation occurred.

transform/utils.go
```diff
@@ -49,7 +49,7 @@ func convert(raw interface{}, jsonType string) (interface{}, error) {
 	if raw == nil {
 		return nil, nil
 	}
-	if rawArray, ok := raw.([]interface{}); ok && len(rawArray) == 1 {
+	if rawArray, ok := raw.([]interface{}); ok && len(rawArray) > 0 && jsonType != "array" {
```
tkuhlman: 👍

jsonschema/validator.go
brianlieberman: yeah that's kinda what I was referring to, it's _most_ likely an array in this instance is really a []byte, but in theory its possible it could get tripped by other things and I had some concern it'd be masking something like you said.  Wrapping and preserving might make the most sense then 👍

transform/transformer.go
```diff
@@ -114,7 +114,7 @@ func (tr *Transformer) jsonTransform(raw json.RawMessage) (json.RawMessage, erro
 
 	valid, err := tr.schema.Validate(out)
 	if err != nil {
-		return nil, fmt.Errorf("transformed result validation error: %v", err)
+		return nil, fmt.Errorf("input sueccessfully transformed but did not match schema: %v", err)
```
zacaudette: `successfully`

jsonschema/validator.go
```diff
@@ -61,6 +61,7 @@ func (v *validator) Validate(raw json.RawMessage) (bool, error) {
 	}
 	if len(result.Errors()) > 0 {
 		sort.Slice(result.Errors(), func(i, j int) bool {
+
```
tkuhlman: nit: extra line added

PR #91, opened 2020-03-20
APIS-4452 Add transform test for html data

zacaudette:


zacaudette:
Currently, this only fixes issues with tests and changes the default json marshaling behavior.

PR #93, opened 2020-04-13
Add option to generate descriptions as comments

tkuhlman:


generate/generate.go
```diff
@@ -108,7 +111,7 @@ func BuildStructsWithArgs(args BuildArgs) error {
 		}
 		embeds = append(embeds, name)
 
-		path, err := buildStructFile(args.SchemaPath, allOfPath, name, packageName, nil, args.OutputDir, args.FieldNameMap)
+		path, err := buildStructFile(allOfPath, name, packageName, nil, args)
```
ahornerr, line 114: :+1:

generate/struct.go
```diff
@@ -26,18 +26,26 @@ type extractedField struct {
...
+func (ef *extractedField) write(w io.Writer, prefix string, required, descriptionAsStructTag bool) error {
 	var omitempty string
 	if !required {
 		omitempty = ",omitempty"
 	}
 	jsonTag := fmt.Sprintf(`json:"%s%s"`, ef.jsonName, omitempty)
 	var description string
-	if ef.description != "" {
+	if descriptionAsStructTag && ef.description != "" {
 		description = fmt.Sprintf(`description:"%s"`, strings.Split(ef.description, "\n")[0])
```
ahornerr, line 37: I know this is existing code but is there any value to not splitting the description on a newline? Seems strange to me that if there's any newlines at all the rest will be omitted

generate/struct_test.go
```diff
@@ -173,17 +173,29 @@ func TestExtractedField_Write(t *testing.T) {
...
 	}{
 		{
 			description: "Write scalr with no prefix",
 			ef:          &extractedField{name: "Field", jsonName: "field", jsonType: "number"},
 			want:        "Field\tfloat64\t`json:\"field,omitempty\"`\n",
 		},
+		{
+			description: "Write scalr with descrition as comment",
+			ef:          &extractedField{name: "Field", jsonName: "field", jsonType: "number", description: "I expect a better description"},
+			want:        "// I expect a better description\nField\tfloat64\t`json:\"field,omitempty\"`\n",
```
ahornerr, line 191: :joy:

generate/struct.go
tkuhlman: It is strange, it took me a couple of tries to get the test files right because of this. If you have a newline in the description it will mess up the go struct that is built so they need to be stripped. I'm guessing when it was first implemented rather than just replacing them with spaces and making a huge description it was considered better to just get the first line. I didn't fix it because I didn't want to change any existing behavior (or really I didn't want to update the test data :)

ahornerr: :joy: Yeah I figured it would break the struct tags but I would much prefer we replace newlines with spaces. I haven't looked into it but how many test files would actually need updating? I don't think we really use that many newlines in descriptions in our actual schema definitions (not test files)

PR #94, opened 2020-04-14
Add a flag to generate Go structs without anonymous nested structs

tkuhlman:


generate/generate.go
```diff
@@ -18,22 +18,31 @@ const disclaimer = "// Code generated by github.com/GannettDigital/jstransform;
...
+//  StructNameMap allows specifying the type name of the struct for each JSON file.
+//
+//  FieldNameMap is used to provide alternate names for fields in the resulting structs.
+//  The property names in the JSON tags for these structs remains the same as supplied.
+//  This can be used to accommodate names that are valid JSON but not valid Go identifiers
 type BuildArgs struct {
 	SchemaPath             string
 	OutputDir              string
 	DescriptionAsStructTag bool
+	NoNestedStructs        bool
```
ahornerr, line 45: Nit: Perhaps we could rename this to `NestedStructs` or `DisableNestedStructs` depending on what you want the default to be. `NoNestedStructs` feels a little clunky to me

generate/test_data/nonest/complex.go
```diff
@@ -0,0 +1,50 @@
...
+
+import "time"
+
+type Complex struct {
+	Simple
+
+	Caption string `json:"caption"`
+	Credit  string `json:"credit"`
+	// The available cropped images
+	Crops          []ComplexCrops      `json:"crops"`
```
ahornerr, line 13: I wonder if we should trim a trailing `s` from the split up struct type name. If you're instantiating an instance of the struct in code (like sports and elections do) then it feels strange to have the trailing `s`. On the other hand, trying to manipulate strings that represent field names can lead us down a bad path, like what I ran into with search GraphQL schemas recently. I'd imagine if we look for a lowercase `s` it would work mostly alright - if we had a field named `SSTS` then it wouldn't remove the last `s`.

main.go
```diff
@@ -82,7 +83,8 @@ func main() {
 		GenerateAvro:           *genAvro,
 		GenerateMessagePack:    *genMessagePack,
 		ImportPath:             *importPath,
-		DescriptionAsStructTag: !*descriptionAsStructTag,
+		DescriptionAsStructTag: *descriptionAsStructTag,
+		NoNestedStructs:        !*nestedStructs,
```
ahornerr, line 87: This looks so weird :joy:

PR #103, opened 2020-06-16
Add support for nullable struct arrays

zacaudette:


generate/avro_helpers.go
```diff
@@ -337,6 +337,9 @@ func (fm *avroFieldMapper) generateStructValue(name, prefix, typeName string, ge
 		} else {
 			structDef = mappedStruct.name
 		}
+		if _, ok := aType.Elt.(*ast.StarExpr); ok {
```
tkuhlman, line 340: Am I missing anything or is this the only non-generated change?

zacaudette: I added the contributors array to the simple schema as a test case for this, otherwise yes the rest are generated changes.

tkuhlman: 👍

PR #104, opened 2020-06-19
Generate non-required time fields as a pointer to allow JSON marshal to omit the fields

tkuhlman:


tkuhlman:
Travis is still having trouble updating the build status, https://travis-ci.org/github/GannettDigital/jstransform/builds/700151507

PR #105, opened 2020-06-25
tests in /generate no longer share test data

jstorer:


ahornerr:
🙅

jstorer:
coverage was checked in goland and this branch mirrors the coverage of the master. It is appears coveralls is having a bad time with it for currently unknown reasons so the value is doing to be tweaked for passing.

PR #108, opened 2020-09-21
APIS-5332 Fix for nested array of structs and testing

mmandolesi-g:


bridgetlane (approved):

generate/avro_test_data/nested.go
```diff
@@ -0,0 +1,14 @@
+package avro_test_data
+
+type Nested struct {
+	FactCheckClaims []struct {
+		Claim          string `json:"claim"`
+		Rating         string `json:"rating"`
+		Author         string `json:"author"`
+		Date           string `json:"date"`
+		AppearanceURLs []struct {
```
tkuhlman, line 9: This file is built with anonymous nested structs. It would be good to have a 2nd file where the nested structs where named. The same schema can generate that if you use the `-nestedStructs=false` flag.

tkuhlman (commented):
The change looks good to me, but I did look for the other named nested structs style of test like Zac recommends and don't see that being run. I expect it will work but I think adding that would be best for completion.

zacaudette (approved):

generate/avro_helpers_test.go
```diff
@@ -52,13 +52,25 @@ func TestBuildAvroHelperFunctions(t *testing.T) {
...
+		{
+			description: "nested array structs",
+			name:        "Nested",
+			filename:    "nested.go",
+			wantPath:    "nested_avro.go",
+		},
+		{
+			description: "nested array structs with named nested structs",
+			name:        "Nested",
+			filename:    "nonest/nested.go",
```
tkuhlman, line 64: "nonest/nested" nice naming :)

tkuhlman (approved):

PR #112, opened 2020-12-11
Setup github actions for testing and remove TravisCI

zacaudette:


.github/workflows/test.yml
```diff
@@ -0,0 +1,36 @@
+name: test
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
tkuhlman, line 7: Lets run it on all pull requests

.github/workflows/test.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-$GITHUB_REPOSITORY-${{ hashFiles('**/go.mod') }}
+      - name: install dependencies
+        run: go mod download
+      - name: go vet
+        run: go vet ./...
+      - name: go test race
+        run: go test -v -race -parallel 4 -coverprofile=covprofile -covermode=atomic ./...
+      - name: install goveralls
```
tkuhlman, line 29: I'm fine with ditching coveralls but that can be discussed and made another PR

.github/workflows/test.yml
```diff
@@ -0,0 +1,36 @@
...
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '^1.15.5'
+      - uses: actions/checkout@v2
+      - uses: actions/cache@v2
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-$GITHUB_REPOSITORY-${{ hashFiles('**/go.mod') }}
+      - name: install dependencies
+        run: go mod download
```
tkuhlman, line 24: I don't think this is needed at all, go test will pull modules it needs.

tkuhlman (commented):
👍🏼 to ditching Travis. Some nice follow ups to this would be automatic tagging.

.github/workflows/test.yml
zacaudette: Agreed, with the mgspack, avro, and some of the generated code there is a lot that won't be tested.

tkuhlman (approved):

PR #113, opened 2021-01-25
APIS-5332 Add removeHTML operation for transforms

mmandolesi-g:


transform/operations.go
```diff
@@ -311,6 +313,25 @@ func (c *toCamelCase) transform(raw interface{}) (interface{}, error) {
...
+// removeHTML is a transformOperation which removes all html from a string.
+type removeHTML struct {
+	args map[string]string
+}
+
+func (c *removeHTML) init(args map[string]string) error {
+	return nil
+}
+
+func (c *removeHTML) transform(raw interface{}) (interface{}, error) {
```
CesarRosasGlobant, line 325: why not use a function that always accepts a string instead of convert the interface{} to string?

mmandolesi-g: That's how it's defined in the interface: https://github.com/GannettDigital/jstransform/blob/9da0e360189861fb47c1bd257367473864203ade/transform/transform.go#L24. Has to accept many different types, not just string.

transform/operations_test.go
```diff
@@ -449,6 +449,23 @@ func TestCurrentTime(t *testing.T) {
...
+			description: "Working case one",
+			in:          "Japanese man calls for a trial by combat when talking to his wife's <br>mother in law<br> &",
+			want:        "Japanese man calls for a trial by combat when talking to his wife's mother in law &",
+		},
+		{
+			description: "Working case two",
+			in:          "<div><br>++Japanese <i>man</i> Collin D'Oliver Murasame <strong>calls for a trial by combat when talking</strong> to his wife<custom>",
+			want:        "++Japanese man Collin D'Oliver Murasame calls for a trial by combat when talking to his wife",
+		},
+	}
```
gannett-ggreer, line 464: There are cases over in https://github.com/GannettDigital/content-api-authoring/pull/262 that failed replacement and would make good unit tests here.

CesarRosasGlobant (approved):
Looks good for me

tomtoday (approved):

PR #114, opened 2021-05-21
APIS-5905_add string to float64 transform

erikturner:


barberta (approved):

transform/transform.go
```diff
@@ -80,6 +80,8 @@ func (ti *transformInstruction) UnmarshalJSON(data []byte) error {
 			op = &toCamelCase{}
 		case "removeHTML":
 			op = &removeHTML{}
+		case "stringToFloat":
+			op = &stringToFloat64{}
```
gannett-ggreer, line 84: Any reason these should be different?  Rest of these are consistent.

transform/operations.go
```diff
@@ -335,6 +335,29 @@ func (c *removeHTML) transform(raw interface{}) (interface{}, error) {
...
+
+func (c *stringToFloat64) transform(raw interface{}) (interface{}, error) {
+	in, ok := raw.(string)
+	if !ok {
+		return nil, errors.New("stringToFloat64 only supports strings")
+	}
+
+	f, err := strconv.ParseFloat(in, 64)
+	if err != nil {
+		return nil, errors.New("error converting string to float64")
```
mmandolesi-g, line 355: You have `err` which is useful to pass up. Use `fmt.Errorf()` and wrap the `err` with some extra context.

PR #115, opened 2021-05-21
Apis 5905_update case statement, change error

erikturner:


transform/operations.go
```diff
@@ -347,12 +347,12 @@ func (c *stringToFloat64) init(args map[string]string) error {
 func (c *stringToFloat64) transform(raw interface{}) (interface{}, error) {
 	in, ok := raw.(string)
 	if !ok {
-		return nil, errors.New("stringToFloat64 only supports strings")
+		return nil, fmt.Errorf("stringToFloat64 only supports strings")
```
gannett-ggreer, line 350: No need to `fmt.Errorf` if you're not using a format (`%`), although I suggest using `%T` here to print out the type that the function did get.

mmandolesi-g: Ya i like printing out the type.

erikturner: updated

transform/operations.go
```diff
@@ -347,12 +347,12 @@ func (c *stringToFloat64) init(args map[string]string) error {
...
 	in, ok := raw.(string)
 	if !ok {
-		return nil, fmt.Errorf("stringToFloat64 only supports strings")
+		return nil, errors.New("stringToFloat64 only supports strings")
 	}
 
 	f, err := strconv.ParseFloat(in, 64)
 	if err != nil {
-		return nil, fmt.Errorf("error converting string to float64: %v", err)
+		return nil, fmt.Errorf("error converting string to float64: %T", err)
```
gannett-ggreer, line 355: You still need `%v` on this one because the `err` isn't a type.  The `%T` was for the first error printing, using `raw` as the argument.

transform/operations.go
```diff
@@ -347,12 +347,12 @@ func (c *stringToFloat64) init(args map[string]string) error {
 func (c *stringToFloat64) transform(raw interface{}) (interface{}, error) {
 	in, ok := raw.(string)
 	if !ok {
-		return nil, fmt.Errorf("stringToFloat64 only supports strings")
+		return nil, errors.New("stringToFloat64 only supports strings")
```
mmandolesi-g: Change this one to `fmt.Errorf("stringToFloat64 only supports strings, raw type: %T, raw")` or something like that.

transform/operations.go
mmandolesi-g, line 355: Keep this one as `fmt.Errorf("error converting string to float64: %v", err)`.

mmandolesi-g (approved):

PR #122, opened 2022-03-03
SDGT-192

jeffbum:
Fix handling of nested array transforms.

A bug has been found in jstransform in which values are not properly accessed when dealing with nested arrays and literal data type outputs instead of objects.

https://jira.gannett.com/browse/SDGT-192


jeffbum:
There is a test failing on master (generate_test.go --> nested.go) due to the description field in the json schema. Is this something you would want me to fix as well in this PR. Or open another ticket ?

transform/transformer_test.go
```diff
@@ -467,7 +487,7 @@ func TestSaveValue(t *testing.T) {
 }
 
 func TestTransformer(t *testing.T) {
-	const parallelRuns = 4
+	const parallelRuns = 1
```
gannett-ggreer, line 490: One of the points of this test is to make sure multiple transforms can run from the same transformer concurrently, so changing this to `1` defeats the purpose of the test.

gannett-ggreer:
The changes to `generate/generate_test_data/generated/nested.go` and `generate/generate_test_data/nonest/nested.go` look unnecessary, if I revert their test's `DescriptionAsStructTag` flag back to what it was before.  Did you have a specific need to change that value in those tests?

transform/transformer_test.go
jeffbum: ah yes. missed this in the post-clean up. will change back.

transform/utils.go
```diff
@@ -159,16 +159,19 @@ func convertString(raw interface{}) (interface{}, error) {
...
+func extractTransformInstructions(raw json.RawMessage, transformIdentifier, path string, instanceType string) (*transformInstructions, error) {
 	rawTransformInstruction, _, _, err := jsonparser.Get(raw, "transform", transformIdentifier)
 	if err != nil && err != jsonparser.KeyPathNotFoundError {
 		return nil, fmt.Errorf("failed to extract raw instance transform: %v", err)
 	} else if len(rawTransformInstruction) == 0 {
 		return nil, nil
 	}
-
 	splits := strings.Split(path, ".")
 	parentPath := strings.Join(splits[:len(splits)-1], ".")
```
DirtySoc, line 170: could move this into an else block after the if to avoid the .Join if not needed.

jeffbum: good call out. done.

transform/utils.go
```diff
@@ -159,16 +159,21 @@ func convertString(raw interface{}) (interface{}, error) {
...
 	}
-
 	splits := strings.Split(path, ".")
-	parentPath := strings.Join(splits[:len(splits)-1], ".")
+	var parentPath string
+	// check if last 3 bytes are "[*] and last 4 are not ".[*]"
+	if instanceType == "scalar" && strings.HasSuffix(path, "[*]") && !strings.HasSuffix(path, ".[*]") {
+		parentPath = path
+	} else {
+		parentPath = strings.Join(splits[:len(splits)-1], ".")
```
gannett-ggreer, line 175: If you're going to move this here then you should move the `splits :=` line too because it isn't used anywhere else but this line.

jeffbum: moved.

transform/utils.go
```diff
@@ -159,16 +159,21 @@ func convertString(raw interface{}) (interface{}, error) {
...
 	if err != nil && err != jsonparser.KeyPathNotFoundError {
 		return nil, fmt.Errorf("failed to extract raw instance transform: %v", err)
 	} else if len(rawTransformInstruction) == 0 {
 		return nil, nil
 	}
-
-	splits := strings.Split(path, ".")
-	parentPath := strings.Join(splits[:len(splits)-1], ".")
+	var parentPath string
+	var splits []string
```
gannett-ggreer, line 170: No need to keep `splits` declared here, just use `splits :=` again in the `else`.

jeffbum: Ah yes. Sorry about that.

gannett-ggreer (approved):

barberta (approved):

PR #123, opened 2022-04-14
SDGT-352 Implement an equivalent function to converToFloat64 that does so for integers

leowotzak:


transform/operations.go
```diff
@@ -15,8 +15,8 @@ import (
 
 var durationRe = regexp.MustCompile(`^([\d]*?):?([\d]*):([\d]*)$`)
 
-// duration is a transformOperation which changes from a string duration like "MM:SS" to a number of seconds as
-// an integer.
+// duration is a transformOperation which changes from a string duration like
+// an integer. "MM:SS" to a number of seconds as
```
gannett-ggreer, line 19: That looks unintentional.

leowotzak: You're right, I must placed it back in the wrong order, will fix

transform/operations.go
```diff
@@ -357,7 +360,36 @@ func (c *convertToFloat64) transform(raw interface{}) (interface{}, error) {
...
+	case int8:
+		return int64(in), nil
+	case int16:
+		return int64(in), nil
+	case int32:
+		return int64(in), nil
+	case int64:
+		return int64(in), nil
+	default:
+		return nil, fmt.Errorf("convertToInt64 only supports strings, int, and float64, raw type: %T", raw)
```
gannett-ggreer, line 387: You don't have `float64` in the list.

leowotzak: I was curious about that. Further raises the question of how to handle the decimal points. Should there be rounding, or should the decimals simply be removed? or should it throw an error?

mmandolesi-g:
You need to add to this switch
```
switch toj.Name {
		case "changeCase":
			op = &changeCase{}
		case "currentTime":
			op = &currentTime{}
		case "duration":
			op = &duration{}
		case "inverse":
			op = &inverse{}
		case "max":
			op = &max{}
		case "replace":
			op = &replace{}
		case "split":
			op = &split{}
		case "timeParse":
			op = &timeParse{}
		case "toCamelCase":
			op = &toCamelCase{}
		case "removeHTML":
			op = &removeHTML{}
		case "convertToFloat64":
			op = &convertToFloat64{}
		default:
			return fmt.Errorf("unsupported operation %q", toj.Name)
		}
```

There's also documentation in `transform.adoc` and `transformSchema.json` to update.

Before merging this i would recommend you run whatever code you're hoping to fix with this locally with this change. Do a `replace` block in the go.mod file pointing to ur local copy from that other repo.

transform/operations.go
gannett-ggreer: When using this kind of function I would expect the intention is for the source to not have a fractional part to begin with.  The caveat is sometimes floating-point can't exactly represent a number so what you think would be an integer isn't due to imprecision, yet int64 is likely big enough to represent it.   So I would just truncate it.

transform/operations.go
leowotzak: https://github.com/GannettDigital/jstransform/pull/123/commits/61184f021faeca8a9ec4a0128dddd935b32a25e3

transform/operations.go
leowotzak: https://github.com/GannettDigital/jstransform/pull/123/commits/8bcccf200d586655e58e513a2707b3a9d8a43004

transform/operations.go
```diff
@@ -357,7 +360,38 @@ func (c *convertToFloat64) transform(raw interface{}) (interface{}, error) {
...
+
+func (c *convertToInt64) init(args map[string]string) error {
+	return nil
+}
+
+func (c *convertToInt64) transform(raw interface{}) (interface{}, error) {
+	switch in := raw.(type) {
+	case string:
+		return strconv.ParseInt(in, 2, 64)
+	case float64:
```
gannett-ggreer, line 376: There is also `float32`, if that matters.

leowotzak: https://github.com/GannettDigital/jstransform/pull/123/commits/3f653fe43b2c06e825cfab875c4033406ebf0ee8

transform.adoc
```diff
@@ -71,5 +71,6 @@ Operations allow further mutation of data, for mutation types that are not curre
 | toCamelCase | string | string | delimiter | The delimiter to split the string on
 | removeHTML | string | string ||
 | convertToFloat64 | string, int, float64 | float64 ||
+| convertToFloat64 | string, int, float64 | int64 ||
```
mmandolesi-g, line 74: Wrong name here.

leowotzak: Good catch, must've snuck into the commit

leowotzak:
I have tested this with `content-api-sport`, and the operation is working successfully.

leowotzak:
Will figure out what issue is

mmandolesi-g:
PR is failing for some reason.

mmandolesi-g (approved):

gannett-ggreer (approved):

PR #125, opened 2022-10-14
SDGT-930 - handle empty XML tags in object XML transformer

liamphmurphy:


gannett-ggreer:
You should add a test and a comment on that new code because its purpose isn't at all obvious.

liamphmurphy:
> You should add a test and a comment on that new code because its purpose isn't at all obvious.

There aren't any tests setup to test the XML transformers in `instance.go` directly, they get tested indirectly from the `transformer_test.go` files. I'll look at adding a few to directly target the XML ones.

gannett-ggreer (approved):

PR #131, opened 2023-06-02
APIS-7217 - Generate GraphQL schemas.

gannett-ggreer:


generate/avro.go
```diff
@@ -392,6 +392,7 @@ func convertToAvroType(cfg avroConfig, expr ast.Expr, name string, nullable bool
 		default: // Another go type
 			n, err := parseGoStruct(t.Name, cfg.dir)
 			if err != nil {
+				//panic(fmt.Errorf("failed to parse Go struct: %v", err))
```
mmandolesi-g, line 395: Can this be deleted?

gannett-ggreer: We could.  I left it because it was a useful debugging tool one time when I was getting `"unknown"` returned unexpectedly.

mmandolesi-g (approved):
This one is probably a good one to walk through with the team.

generate/graphql.go
```diff
@@ -0,0 +1,647 @@
...
+
+	if !filepath.IsAbs(schemaPath) {
+		schemaPath = filepath.Join(filepath.Dir(args.SchemaPath), schemaPath)
+	}
+	schema, err := jsonschema.SchemaFromFile(schemaPath, name)
+	if err != nil {
+		return err
+	}
+
+	// Interface.
```
liamphmurphy, line 94: Not sure I understand what this comment means.

generate/graphql.go
```diff
@@ -0,0 +1,647 @@
...
+				obj.rootStruct.gqlExtractedField.name = obj.rootStruct.gqlExtractedField.jsonName
+				obj.rootStruct.buildType = "interface"
+			}
+		}
+		if err := obj.write(gfile); err != nil {
+			return fmt.Errorf("failed to write allOf struct: %w", err)
+		}
+	}
+
+	// Implementations.
```
liamphmurphy, line 169: Okay, seeing this comment makes the `Interface` comment above make more sense. But I think the comments could still be a bit more descriptive?

gannett-ggreer: Added some more detail to the comments.

gannett-ggreer:
For the case of a schema with 0 AllOf, 1 OneOf, and 0 properties, added a minor fix to not consider the generated type an implementation of an interface (since there is no AllOf).
https://github.com/GannettDigital/jstransform/compare/ccb6952cee16186dc9a654640729dd9adbc347b9..5b7ba582f72720010c8e9d3f17b27e879c3bf731?w=1

liamphmurphy (approved):

PR #134, opened 2023-06-07
APIS-7217 - Handle JSON Schema "type" field optionally being an array in transformer.

gannett-ggreer:


jsonschema/schema.go
```diff
@@ -167,3 +167,40 @@ func mergeProperties(parent, child map[string]json.RawMessage) map[string]json.R
...
+			if adataType != jsonparser.String {
+				return
+			}
+			switch string(avalue) {
+			case "null":
+				nullable = true
+			default:
+				if jsonType != "" {
+					err = fmt.Errorf("union types are not supported: %s", value)
+				}
```
liamphmurphy, line 192: Just confirming that you don't want to just return the err here?

liamphmurphy (approved):

jsonschema/schema.go
gannett-ggreer: Can't return an error there because the `jsonparser.ArrayEach()` calls that as a sub-function, hence the odd dance after it.

PR #135, opened 2023-06-16
APIS-7300 - Add option for generating interface GraphQL implementations in separate files.

gannett-ggreer:


generate/graphql.go
```diff
@@ -200,9 +200,31 @@ func buildGraphQLFile(schemaPath, name, packageName string, args BuildArgs) erro
 			generated.rootStruct.fields[lf] = lfd[lf]
 			generated.rootStruct.fieldOrder = append(generated.rootStruct.fieldOrder, lf)
 		}
+
+		gfile := gfile
```
liamphmurphy, line 204: As a potential nit, should this be `gFile`?

generate/graphql.go
```diff
@@ -200,9 +200,31 @@ func buildGraphQLFile(schemaPath, name, packageName string, args BuildArgs) erro
...
+			if !strings.HasPrefix(oneOfName, name) {
+				outName = name + "_" + outName
+			}
+			outPath = filepath.Join(args.OutputDirGraphQL, outName)
+			gfile, err = os.Create(outPath)
+			if err != nil {
+				return fmt.Errorf("failed to open file %q: %v", outPath, err)
+			}
+			if _, err := fmt.Fprintf(gfile, "# %s\n", disclaimer); err != nil {
+				return fmt.Errorf("failed writing GraphQL: %v", err)
```
liamphmurphy, line 217: `%w` for these errors instead of `%v`?

generate/graphql.go
```diff
@@ -200,9 +200,31 @@ func buildGraphQLFile(schemaPath, name, packageName string, args BuildArgs) erro
...
+			}
+		}
 		if err := generated.write(gfile); err != nil {
 			return fmt.Errorf("failed to write oneOf struct: %v", err)
 		}
+		if args.InterfaceFiles {
+			if err := gfile.Close(); err != nil {
+				return fmt.Errorf("failed to close file %q: %v", outPath, err)
```
liamphmurphy, line 225: Same `%w` here.

generate/graphql.go
gannett-ggreer: Strictly speaking, yes, however that's the name of the variable from the existing (`generate.go`) code so I kept it like that.

generate/graphql.go
gannett-ggreer: These I can at least change to match current proper practices.

PR #140, opened 2023-07-25
SDGT-2086 - add valueExists operation

liamphmurphy:


liamphmurphy:
I'm confused on the test failure... doesn't fail locally, and the got / want in the test looks identical 🤔

transform/operations.go
```diff
@@ -95,6 +96,32 @@ func (c *changeCase) transform(raw interface{}) (interface{}, error) {
 	return nil, errors.New("unknown error in changeCase")
 }
 
+// notEmpty is a transformOperation which returns a boolean depending on if the passed in value is considered "empty", as the definition changes on a per-type basis.
```
adamhostettler, line 99: My only thought is that this might be more straightforward if it were "isEmpty" or "empty" rather than the opposite.

liamphmurphy: Would it make sense to call this `exists` instead? Or some variation, such as `nodeExists`.

adamhostettler: Hm, that could work I think. Might make the most sense considering the different types.

liamphmurphy: Sounds good. One improvement I'd like to make on second thought is to use generics for this (such as allowing it to work on any slice), though I'm struggling to figure out how that'd work here in a clean way. I'll keep messing around a bit.

liamphmurphy: So I've made the change, but it doesn't really seem possible with generics, but only with reflection. Don't know how we feel about using that here, but it does help support many types. Let me know what you think.

adamhostettler: I like the name change, good thought on that.

Reflection.. I'm on the fence. Primarily because we know jstransform can be somewhat intensive with larger schemas so adding a performance hit on top of that may not be great.

liamphmurphy: Yeah I think I agree. It's a little annoying to  have to define `[]xmlquery.Node` when an XML array value is loaded, but since this is loading in from either XML or JSON, the list of types should be fairly small and easily added on.

Leaning toward going back the old way.

adamhostettler: Sounds good 👍 Agree it is tedious to define all the types, though the list shouldn't be too big like you said.

adamhostettler (approved):
Looks good to me

PR #144, opened 2024-01-18
SDGT-7548: Add convertToBool operation

leowotzak:
Converts numbers (across multiple data types), strings, & arrays to boolean values.

Zero/empty values are considered false.

### DIFF
--- a/transform/operations.go
+++ b/transform/operations.go
@@ -450,7 +450,7 @@ func (c *convertToBool) transform(raw interface{}) (interface{}, error) {
 	case int64:
 		return in != 0, nil
 	default:
-		return nil, fmt.Errorf("convertToBool only supports boolean, strings, ints, floats, and array types: %T", raw)
+		return false, fmt.Errorf("convertToBool only supports boolean, strings, ints, floats, and array types: %T", raw)
 	}
 }
 
--- a/transform/operations.go
+++ b/transform/operations.go
@@ -418,6 +418,42 @@ func (c *convertToInt64) transform(raw interface{}) (interface{}, error) {
 	}
 }
 
+// convertToBool is a transformOperation which converts various types to int64.
+type convertToBool struct {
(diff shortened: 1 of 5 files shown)


transform/operations.go
```diff
@@ -418,6 +418,42 @@ func (c *convertToInt64) transform(raw interface{}) (interface{}, error) {
...
+func (c *convertToBool) init(args map[string]string) error {
+	return nil
+}
+
+func (c *convertToBool) transform(raw interface{}) (interface{}, error) {
+	switch in := raw.(type) {
+	case bool:
+		return in, nil
+	case string:
+		return len(in) != 0, nil
```
gannett-ggreer, line 435:
```suggestion
	case string:
		return strconv.ParseBool(in)
```

leowotzak: https://github.com/GannettDigital/jstransform/pull/144/commits/e184684099081c3ed476a3f93e404cf46a81afff

transform/operations_test.go
```diff
@@ -586,6 +586,75 @@ func TestConvertToInt64(t *testing.T) {
...
 
+func TestConvertToBool(t *testing.T) {
+	tests := []opTests{
+		{
+			description: "Working string case",
+			in:          "1",
+			want:        true,
+		},
+		{
+			description: "Failing string case",
```
adamhostettler, line 597: Technically this is a "working" case as it properly transforms, so probably better to name it as such.

leowotzak: You're right "working" is a poor choice of word, I'll replace Working/Failing with True/False

leowotzak: https://github.com/GannettDigital/jstransform/pull/144/commits/5359590e5db0445ebe5d34025a18f4beeec28321

adamhostettler (approved):

PR #146, opened 2024-05-15
Support embedding allOf references as Go model rather than flattening

adamhostettler:
Also addresses:
https://github.com/GannettDigital/jstransform/security/dependabot/6


adamhostettler:
Need to test this thoroughly with graphql generation, as I re-used the Target attribute for now just to get it working.

generate/generate_test_data/targeted.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties": {
+    "crops": {
+      "$ref": "./complex.json#properties/crops",
+      "goModel": "Complex_no_nestedCrops"
```
gannett-ggreer, line 7: Property name + `i.FromRef` should be sufficient to calculate a value rather than being forced to provide a name, which is how the GraphQL-generating code does it.

adamhostettler: So in this case are you thinking of just omitting the goModel part entirely? As in, if a $ref is supplied, then use that model automatically?

Or change the goModel to be a bool.

adamhostettler: Personally I would want the second option so it doesn't force it.

gannett-ggreer: I don't know that there is a downside to always doing it, other than changing the names on existing models (perhaps for the better in cases).  I was thinking the value wouldn't be needed to be supplied in the JSON Schema at all (whether the code still tracks it is a separate question) but I hadn't thought through it enough to say whether we might sometimes want to manually override a model to use an existing structure that it outside the JSON Schema being made.

adamhostettler: Part of the problem with doing it automatically is situations like this:
A shared definitions schema such as the one in sports2, which is not generated into models anywhere, as they are "definitions" and not properties (and not a "type" such as events). So, in that case, schemas that reference those definitions will point to a model that doesn't exist (or at least is not generated automatically).

Just to mention a scenario I came across while testing.

adamhostettler: By "automatically" I mean without the second bool attribute to enable/disable.

gannett-ggreer: The definitions are irrelevant until they're contained in a `$ref` at which point they transparently become part of the schema that is walked to generate the model.  From what I remember of the code, everywhere you're relying on a `GoModel` value from the schema there should already be a `FromRef` value.

adamhostettler: Upon examination, I'm not seeing the information needed to automatically determine the model name.

PR #147, opened 2024-06-20
APIS-7647 transform empty object into map[string]string

miladmikeal:


generate/struct_test.go
```diff
@@ -375,6 +375,7 @@ func TestGoType(t *testing.T) {
 		array       bool
 		required    bool
 		pointers    bool
+		emptyObject bool
```
gannett-ggreer, line 378: I would add `fields` as a test field and then put in your sample string field to the existing test entry so it keeps working and let yours be `nil` and thus 0.

generate/struct.go
```diff
@@ -418,8 +418,12 @@ func (ef *extractedField) goType(required, pointers bool) string {
 			customType = true
 		}
 	case "object":
-		// This only happens with nested structures.
-		goType = "struct"
+		if len(ef.fields) == 0 {
+			goType = "map[string]string"
```
gannett-ggreer, line 422: I'd add a TODO comment about using the `additionalProperties` JSON schema value to more intelligently create the target type in the future.

gannett-ggreer:
When it hits this line:

https://github.com/GannettDigital/jstransform/blob/master/generate/struct.go#L66

you're probably going to get output like:
```
    field map[string]string {
```

adamhostettler:
I think it would be good to see a real-world example using this branch, as a proof of concept. Doing so can also help provide insight into potential bugs.

miladmikeal:
> I think it would be good to see a real-world example using this branch, as a proof of concept. Doing so can also help provide insight into potential bugs.

I plan on testing these changes directly onto my `content-api-shared-datatypes` changes using the replace directive before actually merging.

PR #148, opened 2024-07-08
APIS-7647 allow transform of maps

miladmikeal:


generate/struct.go
```diff
@@ -418,11 +418,23 @@ func (ef *extractedField) goType(required, pointers bool) string {
 			customType = true
 		}
 	case "object":
-		// This only happens with nested structures.
-		goType = "struct"
+		// TODO: use additionalProperties JSON schema value to more intelligently create the target type
+		// when nested struct
+		if len(ef.fields) == 0 || ef.fields == nil {
```
gannett-ggreer, line 423: Don't need the `ef.fields == nil` part because `len(ef.fields) == 0` already covers that.

gannett-ggreer (approved):

PR #151, opened 2024-09-11
APIS-7864 Handle empty properties in graphql as `Any`

mmandolesi-g:


generate/graphql.go
```diff
@@ -475,6 +481,13 @@ func (gof *goGQL) walkFunc(path string, i jsonschema.Instance) error {
 			gqlTypeList = append(gqlTypeList, "null")
 		}
 
+		// Special case for when the properties field of an object is `{}`. Put `Any` instead of the struct name and
+		// don't ignore the empty child struct.
```
gannett-ggreer, line 485: don't ignore?

mmandolesi-g: Removed the `don't`.

gannett-ggreer (approved):

PR #158, opened 2026-08-14
APIS-8789: Migrate jsonparser dependency from GannettDigital fork to upstream latest

eehhgg:


adamhostettler (approved):

transform/transformer.go
```diff
@@ -10,8 +10,8 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/GannettDigital/jsonparser"
 	"github.com/GannettDigital/jstransform/jsonschema"
+	"github.com/buger/jsonparser"
```
mmandolesi-g: Looks like we never setup linting here. This should be down with the non-GannettDigital block below.

eehhgg: Good point. Updated 👍

mmandolesi-g (approved):

